PR #87, opened 2012-06-22
Keyboard normalization

jneen:
This separates out the keyboard handling into `src/key.js`.


laughinghan:
No cause for alarm, just wanted to write this down somewhere, but this PR broke typing in Safari 4.0.5 (which is 2008-2011ish era, so old it's difficult to even obtain to reproduce things; the reason it's 4.0.5 in particular is that that's the only [Multi-Safari](https://michelf.ca/projects/multi-safari/) that will even launch for me). Starting with 7c1a670b68c1ffee8c567e822b5250595efe187c, it's not possible to type more than one character in a row, until you select something first.

PR #112, opened 2012-07-28
Refactor Parser::many and friends to fold rather than yield arrays

laughinghan:
Instead of yielding an array, `Parser::many`, `times`, `atLeast`, and `atMost` take in two additional arguments, the arguments that the `fold` or `reduce` HOF takes: an initial result-so-far (or a function returning one, if it's going to be mutated) and a function that takes in two arguments, the combined/folded/reduced-result-so-far, and the result of this iteration. The final result is the final result-so-far.

`mathSequence`, `MathCommand::parser`, `RootTextBlock::renderLatex`, and even `Parser::atMost` are all simplified by this, while `many` and `times` are just as simple.

### DIFF
--- a/src/latex.js
+++ b/src/latex.js
@@ -5,14 +5,9 @@ var latexMathParser = (function() {
     cmd.adopt(block, 0, 0);
     return block;
   }
-  function joinBlocks(blocks) {
-    var firstBlock = blocks[0] || MathBlock();
-
-    for (var i = 1; i < blocks.length; i += 1) {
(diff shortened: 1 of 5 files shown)


laughinghan:
Closed by GitHub when I deleted the `dev` branch, but we actually do want to close this. The parser combinator framework isn't working out, it can't parse stuff like `\over` and `\choose`, we're looking for something better; worst case scenario, we ~~steal~~ adapt KaTeX's parser.

PR #117, opened 2012-08-10
Environments! ... :'(

laughinghan:
So close...yet so far...

So what I've been thinking is, we've been shoving octogonal, hexagonal, and other assorted increasingly-approaching-square pegs into the round hole that is the double-layered MathQuill tree. Now, it was designed to be super general, so it's probably worth trying in another branch to text to work with it properly, but even at best it'll only be so pretty, and I can't see how we could ever get the semantic tree to work with it. Something we've been talking about doing instead is somehow delegating all the cursor logic to the nearest "environment", whatever that means.

In this branch what I'm aiming to try is to have each environment be directly responsible only for creating nodes of the kind that belong in that environment; then their prototypes have all the logic to deal with them, and in another environment where logic to deal with the cursor and the tree is different, different nodes with prototypes that have different, correct logic are allowed and created.

Something that seems to be working surprisingly well is, for the finite and fairly fixed set of possible cursor operations, movement, selection, and editing, splitting up what logic belongs to the cursor and what belongs to the node: the cursor does some bare-bones logic to determine what node to delegate to, logic that really should be the same in all environments with a tree at all, and when the node does its thing, it calls cursor's methods for stuff like inserting/pre/appending the cursor in places.

Right now, all aspects of cursors' horizontal movement, selecting, and deleting that could be different between math and text, that I can think of without actually implementing a text mode with this cursor, have been delegated to the nodes themselves. They look so similar like they should be sharing some code, actually, but I can't quite figure out a good way.

I also have no idea how to generalize/delegate up/down, seeking (clicking), and `Cursor::selectFrom` (mouse selection).

I think once we have those, though, we can at least get text a lot closer to equal footing with math.

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -18,6 +18,11 @@ var TextBlock = P(Node, function(_, _super) {
       this.replacedText = replacedText;
   };
 
+  _.jQadd = function(jQ) {
+    _super.jQadd.call(this, jQ);
+    this.ch[L].jQize(this.jQ[0].firstChild);
+  };
+
   _.createBefore = function(cursor) {
     var textBlock = this;
     _super.createBefore.call(this, cursor);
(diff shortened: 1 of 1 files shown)


laughinghan:
For some reason GitHub is only showing the last few commits of this PR, when actually the full extent of this was: https://github.com/mathquill/mathquill/compare/d705d0b8b44f8186d1189db5fa46c1f7899c8ee1...8c98c01d0a8da3e48131f9820503300ae77a9491

That's why the PR description appears way out of date compared to the diff GitHub associates with this PR.

PR #238, opened 2014-02-07
Kill jQuery plugin; global MathQuill() returns API obj #big

laughinghan:
## Usage

To create a MathQuill thing, just call `MathQuill.StaticMath()`,
`.MathField()` or `.TextField()` with an HTML element:

``` html
<span id="math-field"></span>
```

``` js
var mathField = MathQuill.MathField($('#math-field')[0]);
assert(mathField instanceof MathQuill.MathField);
assert(mathField instanceof MathQuill.EditableField);
assert(mathField instanceof MathQuill);
```

All of the old `.mathquill('methodName', ...)` methods are available on
this MathQuill object as `.methodName(...)` methods, plus a new `.el()`
method that returns the HTML element:

``` js
mathField.latex('x^2 + y^2 = 1');
$(mathField.el()).detach().appendTo(somewhereElse);
mathField.redraw();
```

The `MathQuill` global API object is also a function that you can use to
get the API object for an HTML element:

``` js
MathQuill($('#math-field')[0]).latex('y = \\pm \\sqrt{ 1 - x^2 }');
assert(MathQuill($('#math-field')[0]) === mathField);
```

---

For now I kept all the old, tired `mathquill-embedded-latex`,
`mathquill-editable`, `mathquill-textbox` CSS class names, but it'd be
nice to use more sensical `mathquill-static-math`,
`mathquill-math-field`, `mathquill-text-field` names instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ so MathQuill defaults to white.)
 
 To hack on MathQuill, you're gonna want to build and test the source files
 you edit. In addition to `make`, MathQuill uses some build tools written on
-[Node][http://nodejs.org/#download], so you will need to install that before
+[Node](http://nodejs.org/#download), so you will need to install that before
 running `make`. (Once it's installed, `make` automatically does `npm install`,
 installing the necessary build tools.)
 


Gregoor:
Hey, awesome work, thanks. Is this version also usable? As it was apparently merged into dev I thought it would be usable but the README still says that jQuery is a dependency..

laughinghan:
Even though MathQuill's API is not as a jQuery plugin, it does unfortunately still require jQuery as a dependency. We do plan to remove that too, though

PR #254, opened 2014-03-04
Further reduce running auto-unitalicization search

laughinghan:
`.sibling{Created,Deleted}(dir{ => , origSibling})`

Add `origSibling` argument to `.sibling{Created,Deleted}()`, so that
`Letter`s can avoid running the auto-unitalicization algorithm in yet
another case: typing or deleting a non-`Letter` at the end of a sequence
of `Letter`s (when not breaking apart or joining together sequences).
`Letter`s now check that if the original and new sibling after the
sibling change are both non-`Letter`s, then no sequences of `Letter`s
could've changed.

For `.siblingCreated()`, this is straightforward, the nodes on either side
of the cursor before inserting stuff were the original siblings of each
other, so just copy the cursor before adopting stuff. (And add an
`origSiblings` argument to `MathElement::finalizeInsert`, and factor
that out as `finalizeInsertingMath()`.)

For `.siblingDeleted()`, this is less straightforward. `.siblingDeleted()`
should only be called by `Controller::deleteDir`, but `Controller::deleteDir`
delegates deleting to `.delete{Towards,OutOf}()` on the nodes themselves,
which may delete as much or as little as they desire, only the nodes
themselves easily know what the original siblings were on either side of
what they choose to delete. So, we now change the contract of
`.delete{Towards,OutOf}()` to, if anything is deleted that its siblings
might want to know about, return a `Fragment` of the disowned nodes.

### DIFF
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -8,7 +8,7 @@
  * Both MathBlock's and MathCommand's descend from it.
  */
 var MathElement = P(Node, function(_, _super) {
-  _.finalizeInsert = function(cursor) {
+  _.finalizeInsert = function(origSiblings, cursor) {
     var self = this;
     self.postOrder('finalizeTree');
     self.postOrder('contactWeld', cursor);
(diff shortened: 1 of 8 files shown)


laughinghan:
Whoops, closed by GitHub when I deleted the `dev` branch, but I still think this was a good idea, re-opened as #511.

PR #269, opened 2014-03-26
Add statelessClipboard option

desa:
The default functionality of paste wraps any string without enclosing
$'s in a latex text command. If `pasteParsedInMathMode` is true any
$-unwapped string will be parsed in the in Math Mode.

Todo: add tests


laughinghan:
Contrary to what GitHub records here, this PR was _not_ merged into `dev`, it was superseded by #285.

PR #280, opened 2014-04-13
Fixed capital letter persistence

d4l3k:
There was an issue with restoring from exported mathquill LaTeX (eg `\R` becomes `\mathbb{R}`). I fixed it by changing the exported symbol. Probably not how it should be fixed. A desired implementation would fix restoring from `\mathbb{}`.

### DIFF
--- a/script/test_server.js
+++ b/script/test_server.js
@@ -6,7 +6,7 @@ var fs = require('fs');
 var child_process = require('child_process');
 
 // constants
-var PORT = +process.env.PORT || 9292;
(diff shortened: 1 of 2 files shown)


laughinghan:
You're right that `\mathbb{}` is horribly broken by being something that is exported but can't be imported. However, I think that your hotfix moves us in the direction of worse rather than better by causing us to export more backslash control sequences that aren't common LaTeX. I think until we have a real solution for `\mathbb{}`, we should just remove these entirely (PR forthcoming).

PR #315, opened 2014-05-30
matrix implementation

andpap:
hi, excellent work,
i see that mathquill have not matrix

1) enter \matrix5x6 for create matrix;
2) mathField.latex('\begin{matrix} 1 & 2 \\ 3 & \sqrt{4} \end{matrix}');


laughinghan:
This was good work, I'm glad this was shared! Some of the compromises chosen, like the `\matrix5x6` syntax, aren't along on the direction we're aiming for, but when matrices become a priority we'll definitely circle back to this.

PR #317, opened 2014-06-05
[dev] Fire MathField:edited handler when swapping between the [< ≤] and [> ≥] pairs

alopatin:


jwmerrill:
This issue is currently affecting Desmos, in the sense that when you change a < to a <=, we don't get the message that we need to reevaluate the expression.

I can't really tell what the conclusion is here. Are you saying that application developers should listen for the input event (or keydown/keypress, etc.) themselves? Or was the point that this PR should be changed to do that?

jwmerrill:
Listening for input events on the element isn't ideal in the case that you have a custom onscreen keypad that can be used to edit mathquill, but the user can also type with a keyboard, because only the latter triggers input events.

I think resolving this issue somehow is really critical, because as a developer, having boxes of math on the page whose value I can track is basically the entire reason for mathquill to exist.

It seems like the right answer is for mathquill to add an `edit/change/input` handler alongside the other handlers like `reflow` and `upOutOf`, but I'm not sure how to go about finding all the places that the handler should be notified.

laughinghan:
Yeah, that's exactly what I think the right answer is. I'm not immediately sure either. I think I was originally (when I last replied to this thread) thinking I'd trigger `edited` (or `input` or `change` or whatever) exactly in `.on('keydown keypress', function(e) { setTimeout(...); });`, but looking back at this it's more useful if MathQuill uses it's privileged information to only trigger `edited` when an edit actually happens, i.e. when the exported LaTeX changes, kinda like how `input` in theory only fires when the text in the `textarea` changes.

I do believe that with MathQuill's current feature set,
- `reflow` should fire in all the situations `edited` should except the one in question ("swapping between the [< ≤] and [> ≥] pairs")
- `edited` should fire in all the situations `reflow` should except calling `.reflow()` via the public API

If so, implementation might be straightforward.

(I am hoping that these will diverge further. More things that don't reflow (ghost parens?), and conversely more auto-detection of reasons reflow besides edits (auto-detecting zooming or font-size changing?), seem like they'd be good things.)

laughinghan:
Wait a minute, I think API calls that write math to the math field should probably also trigger `reflow` but not `edited`, otherwise API calls are difficult to distinguish from actual user input, and you get annoyances like https://github.com/mathquill/mathquill/commit/7d519854d17a391de6a050c978fc55be80234c07#commitcomment-6586893, no? Textareas don't fire `input` if `.value` is changed by JavaScript, only if it changes due to actual user input.

In the case of the custom on-screen keyboard, I think it should be calling the input handler itself.

jwmerrill:
> I think API calls that write math to the math field should probably also trigger reflow but not edited

I was worried you'd say this, and it might be the right call, but from the perspective of a software keypad, it is kind of annoying to draw this distinction. You kind of want consumers of mathquill events to not care about whether the source of the events was typing on a physical keyboard or an implemented-in-js software keypad. And you don't want the software keypad to have direct access to all the consumers of the mathquill events in order to call their callbacks.

Maybe the software keypad should be triggering an "input" event on the mathquill element? Or if not that, what do you think is the right architecture here?

laughinghan:
I don't know. Here's a possibly discriminating question: might these consumers, that don't want to care about whether an event is from a physical or software keyboard, want to be able to distinguish events-due-to-one-of-the-keyboards from events-due-to-the-consumer-themselves-(not the software keyboard implementation)-calling-`.write('some\  latex')`?

If so, then an event that fires whenever MathQuill's LaTeX changes (both from API calls and from user input) still doesn't work, right? What you'd want then is a layer between the higher-level consumer and MathQuill, a wrapper around math fields that attaches a software keyboard and triggers `edited` (or a custom `editedIncludingBySoftwareKeyboard` handler, or whatever), right? That way the consumer can do one thing in their `editedIncludingBySoftwareKeyboard` handler, and another when they themselves call `.write('some\  latex')`.

laughinghan:
This solves a real problem now, we can solve the theoretical problem later (probably with something like [QuillJS' `source`](http://quilljs.com/docs/events/)).

laughinghan:
Merged as #425

PR #325, opened 2014-06-17
Rename 'edited' hook to more accurate 'reflow'

laughinghan:
Also under consideration was `layout`, but that has other meanings
whereas reflow is more specific:
https://github.com/mathquill/mathquill/pull/317#issuecomment-45383191

Just did:

``` sh
sed -i '' s/edited/reflow/g $(git grep -l edited)
```

and rewrote the README entry.

Recommend `--word-diff-regex='\w+|\S'` to view diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,15 +81,15 @@ Any element that has been MathQuill-ified can be reverted:
 MathQuill($('#revert-me')[0]).revert().html(); // => 'some <code>HTML</code>'
 ```
 
-MathQuill has to perform calculations based on computed CSS values. If you
-mathquill-ify an element before inserting into the visible HTML DOM, then once
-it is visible MathQuill will need to recalculate:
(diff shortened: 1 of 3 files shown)


jwmerrill:
Is this now the best way to find out when the result of `mq.latex()` is likely to be different than it was the last time it was called?

The docs don't make it very obvious that `reflow` will definitely be called whenever the latex is edited, if that is in fact the case.

laughinghan:
Fuck I need to fix this. No, there's an edge case where the LaTeX changes but `reflow` isn't called, "when swapping between the [< ≤] and [> ≥] pairs" #317

`reflow` is meant for layout purposes only, and is only guaranteed to be called if the dimensions of the field may have changed. That's kinda dumb, isn't it. I should probably just have an `edited` handler and not micro-optimize recalculating layout.

PR #342, opened 2014-10-24
Adding selectedLatex to public API, a method that returns latex of what ...

hybrisCole:
...you have selected


laughinghan:
Thanks for the new feature! Can you explain how you're using it, so we can have a better idea of the use cases for this feature?

hybrisCole:
It's been a while since I worked on this, but I remember that I needed to do this in order to be able to select text and modify it

PR #354, opened 2014-11-13
Add bower.json to register mathquill on the bower registry

panuhorsmalahti:
bower is a registry for frontend packages. This bower.json makes it possible to add mathquill to bower using the simple instructions explained here:

http://bower.io/docs/creating-packages/

If the maintainers don't want to maintain the bower package, I can also do that.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,19 @@
+{
+  "name": "mathquill",
+  "version": "0.9.3",
+  "main": [
+    "build/mathquill.min.js",
+    "build/mathquill.css"
+  ],
+  "ignore": [
+    "test/**/*"
+  ],
+  "dependencies": {
+    "jquery": ">=1.4.3"
+  },
+  "keywords": [
+    "mathquill",
+    "math",
+    "latex"
+  ]
+}


narcolepticsnowman:
+1

pvenky:
:+1:

jwmerrill:
What is bower's story for representing build information? When you download the mathquill repo (or the releases), there is nothing in the build directory--you have to run "make" to get that. The files listed under "main" in this PR don't exist until you do that. Is that a problem?

I also see that the js and css files are listed under main, but the font file is not. Is that intended, or is that a mistake?

I think the reason this PR hasn't received any attention yet is that none of mathquill's core committers have any experience with bower, so no one feels qualified to review whether this is actually the correct way to set things up.

jwmerrill:
Looks like what jQuery does here is to check built files into the release tags, but not merge those into the main repo: https://github.com/jquery/jquery/commit/4dec426aa2a6cbabb1b064319ba7c272d594a688. That seems like a sensible strategy to me. I think we could probably run `make` and `make basic` and check in the entirety of the `build/` folder.

kschaefe:
@jwmerrill That is exactly what we're doing with our customized version of MathQuill.

Our [v0.9.4-3 release](https://github.com/apollo-ctp/mathquill/tree/v0.9.4-3) does just that sans `make basic`.  I had added some `grunt` tasks to help automate that since there's supposed to be plugins to help push these types of tags to Github, but it doesn't work correctly.  We'll probably back them out. Hand-rolling the release tags isn't too bad.

Also note that we claimed the name `apollo-ctp-mathquill` to leave the raw `mathquill` name for you.

Others (watchers of this issue, etc.) who are interested can grab our release at this time, do note we have several custom options, but everything is currently backward compatible with 0.9.4.

PR #362, opened 2015-01-03
Fix .latex(undefined) throwing TypeError

laughinghan:
Or `.latex()` of anything besides a string, in fact. Now `.latex()` also
coerces to a string, Python-style.


jwmerrill:
Looks good to me.

laughinghan:
woah this is old

jwmerrill:
Yeah, looked like a good change, though. We should do some PR triage soon.

PR #364, opened 2015-01-05
Fix fraction spacing

laughinghan:
Fix fraction-related spacing stuff. The main thing is fixing numerators to not overlap with the fraction bar (which is espesh visible when selecting a numerator), but also slightly improving the spacing of denominators relative to the fraction bar, particularly in SupSubs.


jwmerrill:
Nice work.

PR #365, opened 2015-01-06
Typing ~ should produce \sim

laughinghan:
Note that `~` is a special character in LaTeX, it's a non-breaking space
kinda like `&nbsp;`
http://tex.stackexchange.com/a/41268

### DIFF
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -3,7 +3,6 @@
  ***********************************/
 
 LatexCmds.notin =
-LatexCmds.sim =
 LatexCmds.cong =
 LatexCmds.equiv =
 LatexCmds.oplus =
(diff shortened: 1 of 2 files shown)


src/commands/math/basicSymbols.js
```diff
@@ -190,7 +190,7 @@ LatexCmds.operatorname = P(MathCommand, function(_) {
 
 LatexCmds.f = P(Letter, function(_, super_) {
   _.init = function() {
-    Symbol.p.init.call(this, this.letter = 'f', '<var class="mq-florin">&fnof;</var>');
+    Symbol.p.init.call(this, this.letter = 'f', '<var class="mq-florin">f</var>');
```
jwmerrill: Did you mean to include this change to f/florin?

laughinghan: Fuck. Nope, great catch

src/commands/math/advancedSymbols.js
```diff
@@ -3,7 +3,6 @@
  ***********************************/
 
 LatexCmds.notin =
-LatexCmds.sim =
```
jwmerrill: Does this mean that users of stock mathquill will no longer be able to enter this is \sim?

laughinghan: No, stock MathQuill includes both basic and advanced symbols, `LatexCmds.sim` just got moved to the basic symbols list

PR #368, opened 2015-01-13

kschaefe:


kschaefe:
Huh, guess I don't get pull requests because this now has other changes than my original request, which was supposed to be 3acf78c only.

kschaefe:
Replaced by branch with solo change in https://github.com/mathquill/mathquill/pull/378.

PR #371, opened 2015-01-22
Change zero-width joiner to zero-width space because of wrong

karoltarasiuk:
rendering in Chrome v40.xxxxxx

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -13,7 +13,7 @@ JS environment could actually contain many instances. */
 var Cursor = P(Point, function(_) {
   _.init = function(root) {
     this.parent = this.root = root;
-    var jQ = this.jQ = this._jQ = $('<span class="cursor">&zwj;</span>');
+    var jQ = this.jQ = this._jQ = $('<span class="cursor">&#8203;</span>');
 
     //closured for setInterval
     this.blink = function(){ jQ.toggleClass('blink'); };


karoltarasiuk:
@laughinghan we noticed that in Chrome updated to v40.xx the cursor with &zwj; HTML entity produces thick black box. You can observe it both in

https://www.desmos.com/calculator
http://mathquill.com/

What is the purpose of using zero-width joiner? I noticed that when we replace it to zero-width space, it works fine, but I may not be aware of some other implications.

laughinghan:
Oh thanks! Desmos was just running into this problem today. The zero-width space fix looks awesome, didn't know that existed, we're also looking into getting rid of the hack of having a zero-width character at all

laughinghan:
Actually the [zero-width space](http://en.wikipedia.org/wiki/Zero-width_space) looks so awesome I might have to start using it in other places, like instead of the `<span style="display:inline-block;width:0">&nbsp;</span>`s in `Fraction`s, `SupSub`s, and `Binomial`s! I was just running into a problem where the non-breaking space is sticking out to the right invisibly, increasing the containing `.scrollWidth`

karoltarasiuk:
Yeah, it looked good to me as well. It's just not supported by IE6 or lower, but I don't think anyone would ever care about those anymore. And indeed in some cases it's better than &nbsp;

jwmerrill:
Awesome. Thanks so much for contributing this.

PR #376, opened 2015-02-24
Add 'autoSubscriptNumerals' option

laughinghan:
When enabled, typing `x1` results in `x_1` with the cursor to the right
of the subscript. Typing more numerals extends the subscript without
putting the cursor inside the subscript (thanks `contactWeld`), and
backspacing "peels off" the last symbol in the subscript. Left and right
arrow keys go around the subscript, skipping its contents entirely. The
down arrow still goes into the subscript, but inside the subscript,
typing `_` is ignored, and typing anything other than a symbol (`^` or `/`)
pretends the cursor is to the right of the subscript.

@jwmerrill

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ var mathField = MathQuill.MathField(el[0], {
   sumStartsWithNEquals: true,
   supSubsRequireOperand: true,
   charsThatBreakOutOfSupSub: '+-=<>',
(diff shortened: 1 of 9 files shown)


jwmerrill:
Nice work. This seems like a clean implementation of a kind of complicated feature.

PR #377, opened 2015-02-24
`OMIT_FONT_FACE=true make` omits @font-face{}

laughinghan:
Any non-empty value for $OMIT_FONT_FACE works, e.g. OMIT_FONT_FACE=t or
OMIT_FONT_FACE=1 or OMIT_FONT_FACE=0 or OMIT_FONT_FACE=false.

Applies to any make target that compiles .less to .css, including
`OMIT_FONT_FACE=true make dev` and `OMIT_FONT_FACE=true make basic`.

--

This is an important feature because there's just too many approaches to
@font-face with too many different tradeoffs between performance and
compatibility, there's no one-size-fits-all that MathQuill can use.
Paul Irish maintains what is probably the most comprehensive and widely
cited reference on @font-face techniques, check out all the gotchas even
on the recommended technique among 7 alternatives:

http://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/

Let's not forget the follow-up post that exclusively lists more gotchas:

http://www.paulirish.com/2010/font-face-gotchas/

@jwmerrill

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,9 @@ UGLIFY_OPTS ?= --mangle --compress hoist_vars=true
 
 LESSC ?= ./node_modules/.bin/lessc
 LESS_OPTS ?=
+ifdef OMIT_FONT_FACE
+  LESS_OPTS += --modify-var="omit-font-face=true"
+endif
 
 # Empty target files whose Last Modified timestamps are used to record when
 # something like `npm install` last happened (which, for example, would then be
--- a/src/css/font.less
+++ b/src/css/font.less
@@ -1,10 +1,14 @@
-@font-face {
-  font-family: Symbola;
-  .font-face();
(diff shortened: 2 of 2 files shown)


jwmerrill:
Enjoyed the comment.

Note that you're pinging a github organization or user called font-face, though. You can avoid doing that by writing `@font-face` with the gravè quotes.

laughinghan:
Thanks!

Interestingly, GitHub appears to have anticipated that problem, @font-face appears to be a placeholder GitHub organization.

Also, first time I've heard someone call backticks "gravè quotes", didn't know what the gravè accent accent was called until now.

jwmerrill:
I checked after I wrote that and couldn't find anything to back up that
diacritic above the e. I guess they're just called grave symbols.

PR #378, opened 2015-02-24

kschaefe:


laughinghan:
Whoops, closed by GitHub when I deleted the `dev` branch, but this was a good idea, re-opened as #510.

PR #379, opened 2015-02-24
Fix text() error when editable field contains '\'

kschaefe:
If you called text() on an editable field where the user has entered a
backslash but hasn’t completed a LaTeX template, the backslash
command’s text() method would throw an error when trying to call text()
on the MathBlock’s undefined ends[L]. Subsequent keystrokes would then
occur outside the template.

This commit ensures that the text() method works when an editable field
contains a backslash that hasn’t been converted into a LaTeX template.
If the editable field contains ‘\’ or part of a LaTeX command like
‘\sqr’, text() will return what the user has typed literally (‘\’ and
‘\sqr’, respectively).

Added units tests for both valid and invalid LaTeX templates.  Unit
tests should fail on any environment that does not merge the math.js
changes.

### DIFF
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -357,9 +357,9 @@ var MathBlock = P(MathElement, function(_, super_) {
   _.html = function() { return this.join('html'); };
   _.latex = function() { return this.join('latex'); };
   _.text = function() {
-    return this.ends[L] === this.ends[R] ?
+    return (this.ends[L] === this.ends[R] && this.ends[L] !== 0) ?
       this.ends[L].text() :
-      '(' + this.join('text') + ')'
+      this.join('text')
     ;
   };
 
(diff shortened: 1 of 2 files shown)


laughinghan:
`.text()` is really unfortunate because there's no spec or rationale for how it should work, but erroring is super bad, thanks for finding and fixing this!

laughinghan:
I just want to say that, just like your latest PR fixing the InterVer typo, this was an exemplary PR in all regards&mdash;minimal code change, added tests, detailed description&mdash;and I'm ashamed it took me as long as it did to merge it. I'm intent on doing better though, I hope you'll keep contributing such good work!

Pankaj1008:
enter button on keyboard work with 
handlers: {
                  enter: function() { 
                   $scope.answerMathField.write('\\textcolor{}{\\text{ }}');
               }
            }
but backspace button not working properly, need to press three, four time to back to previous line

PR #387, opened 2015-04-10
Convert uses of &nbsp; to zero width space.

jwmerrill:
`&#8203;` is the zero width space character. Even with width set to 0, `&nbsp;`
seems to overflow its container in some circumstances. &#8203; is more reliably
zero width.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -240,7 +240,7 @@ var SupSub = P(MathCommand, function(_, super_) {
       block.adopt(this, 0, this.sup).upOutOf = this.sup;
       block.jQ = $('<span class="mq-sub"></span>').append(block.jQ.children())
         .attr(mqBlockId, block.id).appendTo(this.jQ.removeClass('mq-sup-only'));
-      this.jQ.append('<span style="display:inline-block;width:0">&nbsp;</span>');
+      this.jQ.append('<span style="display:inline-block;width:0">&#8203;</span>');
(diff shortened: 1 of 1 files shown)


laughinghan:
More specifically, the `&nbsp;` always overflows the containing span, because the default value of `overflow` is `visible`. However, this is only noticeable with a subscript at the end of the math field because then the `scrollWidth` of the math field (which is `overflow: hidden`) expands to include the non-breaking space, but the actual width of the math field was computed by the browser based on the 0 width of the span containing the non-breaking space, and so we "pan"/"scroll" to the right a bit to "show" the overflowing non-breaking space, causing all the math contents to shift to the left slightly but noticeably.

laughinghan:
We're already relying on zero-width space character support in the cursor now, so I'm gonna merge this for now, but if we ever circle back to this, other solutions to consider:
- `overflow: hidden` on the span should keep the `&nbsp;` from overflowing (it also could be something visible like `x` then)
- it might be possible to get rid of the non-semantic extra span and text contents in the first place, and use `vertical-align: text-bottom` on superscripts and numerators so they'll be vertically aligned correctly?  (Or `vertical-align: bottom`? Or `overflow: hidden` as a hack so that ["the baseline is the bottom margin edge"](http://www.w3.org/TR/CSS21/visudet.html#propdef-vertical-align) (completely different from the above uses of/bugs related to `overflow`))

PR #391, opened 2015-05-12
Feature: restrictMismatchedBrackets

laughinghan:
Say you want to type "(|x|+1)", LaTeX: `\left( \left|x\right| + 1 \right)`
(whitespace added for clarity).

With the current auto-expanding parens behavior (#262 and #281), if you
tried to type that directly, you'd get: `\left(\right| x \left|+1\right)`
(whitespace added for clarity).

This is unintuitive partly because in practice you almost never actually
want mismatched brackets, you actually almost always want pairs of them
to be matching. In fact, usually the only mismatched pairs you want are
like [a,b) and (a,b].

The `restrictMismatchedBrackets` option restricts the mismatched pairs
that may be typed using auto-expanding parens and ghosts to just those
two, otherwise you can only close an auto-expanded pair/solidify a ghost
paren by typing the opposing delimiter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ var el = $('<span>x^2</span>').appendTo('body');
 var mathField = MathQuill.MathField(el[0], {
   spaceBehavesLikeTab: true,
   leftRightIntoCmdGoes: 'up',
+  restrictMismatchedBrackets: true,
   sumStartsWithNEquals: true,
   supSubsRequireOperand: true,
   charsThatBreakOutOfSupSub: '+-=<>',
@@ -175,6 +176,11 @@ can't get to it with just Left and Right, you have to press Down); which is
 the same behavior as the Desmos calculator. `'down'` instead means it is the
 numerator that is always skipped, which is the same behavior as the Mac OS X
 built-in app Grapher.
+
+If `restrictMismatchedBrackets` is true then you can type [a,b) and [a,b), but
+if you try typing `[x}` or `\langle x|`, you'll get `[{x}]` or
+`\langle|x|\rangle` instead. This lets you type `(|x|+1)` normally; otherwise,
+you'd get `\left( \right| x \left| + 1 \right)`.
 
 If `sumStartsWithNEquals` is true then when you type `\sum`, `\prod`, or
 `\coprod`, the lower limit starts out with `n=`, e.g. you get the LaTeX
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -521,8 +521,8 @@ var Bracket = P(P(MathCommand, DelimsMixin), function(_, super_) {
     return '\\left'+this.sides[L].ctrlSeq+this.ends[L].latex()+'\\right'+this.sides[R].ctrlSeq;
   };
   _.oppBrack = function(node, expectedSide) {
-    // node must be 1-sided bracket of expected side (if any, may be undefined),
-    // and unless I'm a pipe, node and I must be opposite-facing sides
+    // return node iff it's a 1-sided bracket of expected side (if any, may be
+    // undefined), and of opposite side from me if I'm not a pipe
     return node instanceof Bracket && node.side && node.side !== -expectedSide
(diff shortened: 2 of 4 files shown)


laughinghan:
@jwmerrill: I remembered another major use case for mismatched brackets that I can't believe you didn't think of: [bra-ket notation](http://en.wikipedia.org/wiki/Bra%E2%80%93ket_notation), man.

On the other hand, also thought of a major counter-use-case-whatever-thingy: set-builder notation, when you want a solitary unpaired pipe inside paired curly braces. (Which is actually currently still achievable in MathQuill with `\vert`, but that's an accident. Also, naturally LaTeX has lots more alternatives to just pipe `|` and `\vert`, there's `\mid`, `\divides`, `\lvert`, `\rvert`, and more: http://tex.stackexchange.com/a/506)

jwmerrill:
Maybe this is crazy, but could we re-associate if you go from `bab` to `baab`; i.e. you start off with `(ba)b` and when you type the second b, change to `(b(aa)b)`. In case my notation isn't clear, `a` and `b` are supposed to be different kinds of braces, and the parens show how they are nested into left/right variants.

jwmerrill:
In bra-ket notation, you don't really want to think of the pipe as matching/closing the `\langle`. It's still the `\rangle` that closes the `\langle`.

jwmerrill:
It strikes me that there are actually two separate cases here

1. The open/closed set notation, where a bracket is matched with a bracket that would not normally be its partner, e.g. the `(` and `]` in `(2,4]` form a pair.

2. Pipes are sometimes used without a partner at all, e.g. in set builder notation, `{` and `}` form a pair, and the pipe doesn't have a partner at all, e.g. in `\{y | f(y) = 0\}`.

Bra-ket notation actually straddles the two cases; ` \hat{p} | \psi \rangle` is case 1, and `\langle \phi | \psi \rangle` is case 2. I'm not 100% sure how to classify `\langle \psi | \hat{p} | \psi \rangle`

I think we should maybe hold off on merging this PR until the ideas settle down, because I think it would be better to ship with it being difficult to type `(|x| + 1)^2` than to add an option to the mathquill API that we'll want to change once we think about it more.

laughinghan:
Re `baab`: you mean when you type a second `a`, not a second `b`, right? That means the first `a` switches directions, right? Only the pipe does that though, no other delimiter does that (I think).

Re bra-ket: you're right, I forgot bra-ket also can have the set-builder notation problem of a singular pipe.

I actually don't feel like we need to be particularly cautious about this PR, it's more UI stability not really API stability, in relatively edgy cases; and the `(|x|+1)^2` situation seems like a really bad experience, and [a,b) and (a,b] are really the only important things we need to enable.

I was also considering making `restrictMismatchedBrackets` take in a string (or maybe list of strings), not a boolean, of the form `'[) (] \\langle| |\\rangle'`, i.e. a space-delimited list of pairs of delimiters. A conservative option, if we want API extensibility but also don't want to add this ill-thought-out feature and have to support arbitrary pairs of delimiters forever, is to say `restrictMismatchedBrackets` must be a string, but it may only have the value `'[) (]'` (or `undefined` to override).

kschaefe:
You also need unmatched curly brackets for piece-wise functions.  Though it doesn't appear that MQ currently supports any of the LaTeX mechanisms to laying out the piece-wise functions themselves.

jwmerrill:
Talked to Han about this some more, and I'm now convinced that `restrictMismatchedBrackets` will get the job done well enough for users of mathquill "basic math" almost all of the time. We also discussed explicitly specifying the pairs of mismatched brackets that you want to allow, but it doesn't seem worth the added implementation complexity at this point.

Code and tests look good to me, so I'm going to ahead and merge this.

laughinghan:
@kschaefe I figure we would directly implement `\begin{case}...\end{case}` rather than bothering getting `\left\{...\right.` to work, do you really think the latter is necessary?

kschaefe:
No.  I didn't realize that 'case' was on the list of things to implement, that would be much better.

PR #392, opened 2015-05-12
Stop using florin, and just special case spacing of the f character

jwmerrill:
Mathquill has been rewriting italicized f's to florin characters because italicized f's take up a lot of horizontal space and collide with parens if you don't treat them specially. This PR switches back to using f's, but gives them a little bit of extra horizontal space to avoid collisions.

The rendering in this branch is actually closer to the way that italic f's are rendered in latex computer modern. Compare:

Florins (mathquill's current strategy):
image

F's with no special case spacing:
image

F's with special case spacing (this PR):
image

Latex with computer modern:
image

For comparison, here are a few h's in a row in latex (notice the relatively tighter spacing):
image

My proximate motivation for doing this is that I want f's to not show up italicized in mathrm mode, and this PR also solves that problem, but I think it's a good idea independent of that issue.

### DIFF
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -398,7 +398,8 @@ var MathBlock = P(MathElement, function(_, super_) {
   };
   _.chToCmd = function(ch) {
     var cons;
-    if (ch.match(/^[a-eg-zA-Z]$/)) //exclude f because want florin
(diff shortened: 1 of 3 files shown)


laughinghan:
This awesome. Way less hacky and closer to the eventual solution of our font just being kerned right.

PR #398, opened 2015-05-19
Don't automatically subscript after functions

### LINKED ISSUE #390: In autoSubscriptNumerals mode, don't autosubscript unitalicized commands
The rationale here is that typing `sin90` is way more likely to mean `sin 90` than `sin_{90}`.

eluberoff:
Fixes #390 
Adds tests for auto-subscripting behavior


jwmerrill:
Looks good to me.

laughinghan:
This is awesome, thanks @eluberoff! I can't believe I didn't write tests for this initially

Issue #396, opened 2015-05-15
option: deleting one parenthesis deletes both


eluberoff:
I absolutely love ghost parens, but the behavior when deleting doesn't feel quite right. Specific case that keeps throwing me off:

(1) type `sin(x)`
(2) delete that first paren after `sin`

Expected: parens cleared.
Observed: (sin x) (with the first paren as a ghost paren)

This definitely feels wrong to me. I think clearing out parens would feel better to me. It's possible that this is only true when deleting from inside the parens -- so backspacing the opening paren clears parens, or deleting the closing paren. For simplicity, I'd prefer either an option (or default) that deleting either hardened paren deletes the other one also.

laughinghan:
Didn't @jwmerrill deal with this in Desmos by having the `sin()` button actually just type `sin(`?

The thing about this is that it's path-dependent, right? If you have `a+b` and you're trying to wrap the `b` in parens and you write the close-paren first, so with the ghost it's `(a+b)`, and then you try to type an open-paren but mistakenly type it before the plus sign so you have `a(+b)`, and you backspace over the open-paren, you don't want to clear the close-paren, you want it to stay there so you can move over and write the open paren and get `a+(b)`, right?

Also, for what it's worth, maybe you've just gotten used to MathQuill's weird behavior? In normal text, you expect to have to backspace over both parens, right?

eluberoff:
I think the case you describe is pretty edge case. Basically, I think that folks (well, I) more often delete a paren because I want to get rid of them than because I want to move one, especially since ghost parens are so freaking awesome/intuitive for entering parens initially.

So, deleting both (current master-branch mathquill behavior) results in (excluding moving the cursor):
2 strokes -> 3 strokes for moving a parenthesis (remove them, add 2 new ones with sweet ghost parens)
2 strokes -> 1 stroke for removing parentheses

IMO, worth the trade-off. If there's disagreement on that, a flag seems reasonable. I'm pretty sure it's what Desmos will want to use.

eluberoff:
oh and jwmerrill's fix is a good one, but there are a ton of other ways for you to end up with parens that you'd like to clear. another one: lots of people type parens to form fractions because they don't know about selecting and then dividing. So you end up with (1+2)/3 and then want to clear off the parens. They harden immediately on dividing, so peeling the parens is more of an ordeal if you need to delete both sides independently.

jwmerrill:
Let's try to come to a consensus on the best behavior here before we start talking about making an option.

[...]

PR #399, opened 2015-05-19
Delete both parens when they match and are both hardened

eluberoff:
Proposed solution for #396 

Leaves current behavior for mismatched parens (power user stuff in there). Leaves current behavior if either is a ghost.

Note: @laughinghan paren behavior is absurdly well tested. well-done / holy crap.


eluberoff:
closing in favor of #400

PR #400, opened 2015-05-20
Unwrap parens when you delete the left paren

### LINKED ISSUE #396: option: deleting one parenthesis deletes both

eluberoff:
Alternative to #399 
Fixes #396 

Tried out deleting both when you delete either, but that doesn't actually feel quite right. Realized that a possible general principle is: actions shouldn't cause non-local changes to the left of the cursor whenever possible. That's what feels weird to me.

Note: this still doesn't help with the case of using parentheses to form a complex fraction. Making a ticket about that one separately.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -578,11 +578,7 @@ var Bracket = P(P(MathCommand, DelimsMixin), function(_, super_) {
     if (
       //unwrap if deleting either...
       side === this.side || //the solid side with a ghost pair
-      (
-        this.side === 0 && //or one side of matched parens
-        side === -1 &&
-        this.sides[1].ch === OPP_BRACKS[this.sides[-1].ch]
-      )
+      (this.side === 0 && side === -1) //opening paren
     ) {
       this.unwrap();
       sib ? cursor.insDirOf(-side, sib) : cursor.insAtDirEnd(side, parent);
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -575,7 +575,15 @@ var Bracket = P(P(MathCommand, DelimsMixin), function(_, super_) {
   _.deleteSide = function(side, outward, cursor) {
     var parent = this.parent, sib = this[side], farEnd = parent.ends[side];
 
-    if (side === this.side) { // deleting non-ghost of one-sided bracket, unwrap
+    if (
+      //unwrap if deleting either...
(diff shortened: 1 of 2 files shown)


jwmerrill:
Why did you close this? Can you reopen it?

eluberoff:
note: you can play with this behavior at desmos.com/calculator

laughinghan:
I think the the asymmetrical left-right behavior is pretty weird. Do you feel strongly about your rule applying to deleting forwards (`Del`)? What if the rule was instead deleting outwards vs inwards, which matches up exactly with your rule when using Backspace?

Specifically, deleting outwards (backspacing the left-paren, or deleting forwards over the right-paren) unwraps parens, whereas deleting inwards ghostifies the paren? I like this because it becomes perfectly consistent with what #411 would do to all other commands: typing `(x+1)^2` is 7 keystrokes, and it takes exactly 7 backspaces to delete it; but if you backspace from the beginning of the exponent or paren, it unwraps the `2` or `x+1`, respectively.

I guess we could apply your rule to get that behavior too; what would delete forwards out of the right end of the exponent do, just move the cursor right?

jwmerrill:
Eli and I discussed this in the office a little while ago. Neither of us
uses forward delete ever, so if you have some intuition about what should
happen in that case, let's go with it.

eluberoff:
+1. outward vs inward seems reasonable.

laughinghan:
I actually want to slightly restrict when unwrapping happens. DCG currently unwraps parens even if there's a half-open one that could be closed, for example, here this ghost close-paren stays a ghost:
unwrap even if could match up

As opposed to not unwrapping and letting them matching up
match up instead of unwrap

@eluberoff: what do you think? https://github.com/laughinghan/mathquill/tree/feature.backspacing-left-paren-unwraps

PR #402, opened 2015-05-21
Denominators should have same padding as numerators

laughinghan:
Before:
image
After:
image

Numerators have a .1em padding on the left and right, but it was removed
from denominators in #397 because the negative margin was causing
problems. Well, florin doesn't use negative margins anymore and neither
should denominators.


jwmerrill:
That's a lot better. Nice change.

PR #403, opened 2015-05-21

laughinghan:

[2]: https://developers.google.com/mobile/articles/fast_buttons

--

tap highlight box.

### DIFF
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -94,6 +94,15 @@ var AbstractMathQuill = P(function(_) {
     this.__controller.root.postOrder('reflow');
     return this;
   };
+  _.fastClick = function(target, clientX, clientY) {
+    var ctrlr = this.__controller, root = ctrlr.root;
+    var el = document.elementFromPoint(clientX, clientY);
+    if (!jQuery.contains(root.jQ[0], el)) el = root.jQ[0];
+    ctrlr.seek($(el), clientX + pageXOffset, clientY + pageYOffset);
+  };
+  _.ignoreNextMousedown = function(fn) {
+    this.__controller.cursor.options.ignoreNextMousedown = fn;
+  };
 });
 MathQuill.prototype = AbstractMathQuill.prototype;
 
--- a/src/services/mouse.js
+++ b/src/services/mouse.js
@@ -3,6 +3,7 @@
  *******************************************************/
 
 Controller.open(function(_) {
+  Options.p.ignoreNextMousedown = noop;
   _.delegateMouseEvents = function() {
     var ultimateRootjQ = this.root.jQ;
     //drag-to-select event handling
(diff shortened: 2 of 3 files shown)


laughinghan:
I'm pretty torn about the name for this API method. `.touchTap()` is what it was called when I first implemented this in a the proprietary Desmos branch, but there it also made the cursor handle show up; that was not enough control over the cursor handle, and if I implement that again I'd certainly make it a wholly separate API method. So here I renamed it `.fastClick()` thinking that it's meant to be used by [FastClick]/["Fast Button"]-style code; but now I'm thinking of going back, because until someone reads the description and those articles, `.fastClick()` would be a completely indecipherable name.

[FastClick]: https://github.com/ftlabs/fastclick
["Fast Buttons"]: https://developers.google.com/mobile/articles/fast_buttons

What else is there? `.fastTap()`? `.fastTouchTap()`? I kinda would like to capture the idea that the API consumer is informing MathQuill of a logical touch tap; `.sendTouchTap()`? `.doTouchTap()`? `.doFastClick()`? `.doClick()`. It def can't just be `.click()` because there's a DOM method named that, and its similar in that it simulates a click, but its signature is totally different, it doesn't take in any arguments, whereas this has to take in coords to seek the cursor towards.

jwmerrill:
This looks good to me. I think `fastClick` is a fine name, and I don't think any of the other proposals are better than it. As Eli mentioned, I think it'd make sense for a fastClick to cause a focus if the element isn't already focused.

jwmerrill:
Actually, `fastTouchTap` might be a better name, since it makes it obvious that this is about touch.

src/publicapi.js
```diff
@@ -94,6 +94,15 @@ var AbstractMathQuill = P(function(_) {
     this.__controller.root.postOrder('reflow');
     return this;
   };
+  _.fastClick = function(target, clientX, clientY) {
+    var ctrlr = this.__controller, root = ctrlr.root;
+    var el = document.elementFromPoint(clientX, clientY);
+    if (!jQuery.contains(root.jQ[0], el)) el = root.jQ[0];
+    ctrlr.seek($(el), clientX + pageXOffset, clientY + pageYOffset);
```
jwmerrill: Do you want this and ignoreNextMousedown to be chainable? Should they `return this`?

laughinghan: they absolutely should, I'm an idiot

laughinghan:
I like `fastTouchTap` better for that reason too, the only thing is that no one else calls it that

laughinghan:
Another direction is to name it for what it does, not when it should be called, like `.simulateClick()` or `.clickAt()` (~~or `.simClick()` or `.execClick()` or `.executeClick()`~~ these are ugly). Note that just `.click()` is bad because there's a DOM method of that name, with a different signature (no arguments, hence `.clickAt(clientX, clientY)` to differentiate).

kschaefe:
How does this change interact with other fast-click libraries, such as [jQuery UI Touch Punch](http://touchpunch.furf.com/)?

jwmerrill:
Bump.

laughinghan:
Whoops, closed by GitHub when I deleted the dev branch, but this was really good work, re-opened as #513.

PR #404, opened 2015-05-22
Use box-sizing: border-box

eluberoff:


laughinghan:
Apparently MathQuill has already been relying on `box-sizing: border-box` since #261 (3f963a445a303b17699ad3d45f2b4f98105385c9), we should totally just go all the way, make stuff like fractions cleaner.

Issue #405, opened 2015-05-23
[dev] Setting charsThatBreakOutOfSupSub = '+-' makes it hard to type 10^{-6}


jwmerrill:
You end up with `10^{}-6`.

I think we might need to special case +/- being used as unary operators, and not break out of exponents in that case.

jwmerrill:
Re: special casing--maybe we can get away with not doing any special casing, and just say that nothing breaks out of exponents when it's used as the first character. This is the rule currently being used in production Desmos, and it seems to work out fine.

It results in `10^--6` rendering as `10^{-}-6`, which isn't ideal, but I'm not sure that's a case we need to optimize for.

laughinghan:
Sounds good for now.

PR #407, opened 2015-05-23
Unary operators shouldn't break out of exponents.

jwmerrill:
The point of this is to make it so that even with `charsThatBreakOutOfSupSub = '+-'`, typing `10^-7` still gives you `10^{-7}`.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -191,12 +191,8 @@ var SupSub = P(MathCommand, function(_, super_) {
         else cursor.clearSelection().insRightOf(this.parent);
         return cmd.createLeftOf(cursor.show());
       }
-      if (
-        cursor.options.charsThatBreakOutOfSupSub.indexOf(ch) > -1 &&
-        // Unary operators never break out of exponents, so that it's easy to
-        // write negative exponents.
-        cursor[L] !== 0
-      ) {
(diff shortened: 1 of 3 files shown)


src/commands/math/commands.js
```diff
@@ -191,7 +191,12 @@ var SupSub = P(MathCommand, function(_, super_) {
         else cursor.clearSelection().insRightOf(this.parent);
         return cmd.createLeftOf(cursor.show());
       }
-      if (cursor.options.charsThatBreakOutOfSupSub.indexOf(ch) > -1) {
+      if (
+        cursor.options.charsThatBreakOutOfSupSub.indexOf(ch) > -1 &&
+        // Unary operators never break out of exponents, so that it's easy to
+        // write negative exponents.
+        cursor[L] !== 0
```
jwmerrill: Is there a better way to check whether an operator is being used as a unary prefix operator than to check whether the cursor is at the beginning of the block?

jwmerrill:
Fixes #405.

src/commands/math/commands.js
laughinghan: I can't think of anything that would generalize to, like, `\sim`, if that's what you're hoping. I think this is good. I updated the README too

PR #408, opened 2015-05-24
Replace zero width joiner with zero width space.

jwmerrill:
`&#8203` is the zero width space character. We replaced most other uses of the zero width joiner with a zero width space because of a font rendering bug in Chrome. For whatever reason, this particular example doesn't visually exhibit the bug, but I think it makes sense to replace zwj here for consistency, and because a zero width space makes more semantic sense.

### DIFF
--- a/src/cursor.js
+++ b/src/cursor.js
@@ -15,7 +15,7 @@ var Cursor = P(Point, function(_) {
     this.parent = initParent;
     this.options = options;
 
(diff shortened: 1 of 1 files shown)


laughinghan:
Wait a minute this is...the same spot...where we replaced the zero-width joiner the first time: #371

I guess we forgot to merge `master` back into `dev`, and before we looked again, Chrome had fixed their wack rendering of the `&zwj;` character. Just merged `master` back into `dev`.

PR #410, opened 2015-05-27
check that we have a select method before using it

eluberoff:
textareas have a select method but spans do not. check for the existence of this method before executing it. fixes non-user-visible thrown error on iPads in the Desmos Calculator


laughinghan:
Lovely. Great catch. Adding test case and merging.

PR #411, opened 2015-05-27
Simplify backspace behavior

eluberoff:
On this branch, backspace into a complex element (fraction, exponent, etc) behaves like a left arrow instead of selecting.

to me, it's a little more intuitive. A few edge cases feel a little off, but the whole thing feels more intuitive to me. @laughinghan, what do you think?

Also adding a bunch of backspace tests -- backspace felt pretty untested

### DIFF
--- a/test/unit/autosubscript.test.js
+++ b/test/unit/autosubscript.test.js
@@ -2,6 +2,9 @@ suite('autoSubscript', function() {
   var mq;
   setup(function() {
     mq = MathQuill.MathField($('<span></span>').appendTo('#mock')[0], {autoSubscriptNumerals: true});
+    rootBlock = mq.__controller.root;
+    controller = mq.__controller;
+    cursor = controller.cursor;
   });
   teardown(function() {
     $(mq.el()).remove();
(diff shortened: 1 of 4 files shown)


jwmerrill:
Code looks good to me, and I like this new behavior.

laughinghan:
Interesting that you ignore `leftRightIntoCmdGoes`, so that on Desmos right now, left into a fraction goes into the numerator, but backspacing into a fraction goes into the denominator?

laughinghan:
`Ctrl-Backspace` is weird now, it ends up deleting everything up until the beginning of some inner commands, like from
image
to
image
Trying to fix right now

laughinghan:
Oh I didn't answer your question. I really like this behavior! I think I floated a similar idea last year, suggesting that it felt weird that with the selection-when-you-backspace behavior, huge fractions and tiny fractions are both deleted with two backspaces; it seems desirable for the number of backspaces it takes to delete something, to scale with the number of keystrokes it took to type it.

(Though for what it's worth, the Mac OS X built-in Grapher app has the selection-when-you-backspace behavior, but Jeanine, who's a Linux user and has certainly never used Grapher, suggested it independently, so there's something intuitive about that, too.)

jwmerrill:
One thing that's bugging me about the new behavior is when I have an empty exponent block, it feels weird that backspace takes me into it instead of getting rid of it. Once I'm inside the exponent that has nothing in it, it doesn't feel obvious that the correct thing to do is press "backspace" again to actually get rid of the block.

test/unit/autosubscript.test.js
```diff
@@ -46,4 +49,58 @@ suite('autoSubscript', function() {
   });
 
 
+  test('backspace through compound subscript', function() {
+    mq.latex('x_{2_2}');
+
+    //first backspace moves to cursor in subscript and peels it off
```
laughinghan: love these comments, they make the tests so much more readable

PR #413, opened 2015-05-29
Fix tabbing with selection

eluberoff:


jwmerrill:
LGTM

src/services/keystroke.js
```diff
@@ -158,6 +158,13 @@ Controller.open(function(_) {
     prayDirection(dir);
     var cursor = this.cursor;
 
+    //tabbing out of selection behaves like pressing the relevant arrow direction
+    if (cursor.selection) {
+      if (dir === R) this.moveRight();
+      if (dir === L) this.moveLeft();
```
laughinghan: `this.moveDir(dir);` :smile:

laughinghan:
Nice catch, I hadn't noticed the current broken behavior. Do you feel strongly about this behavior rather than tabbing to another field even if something's currently selected, like back on 561b1f0098c0770d5d7ff2491ae5298df641d00a before #294 caused the current broken behavior? I liked the invariant that whenever the cursor (or selection) is in the root, the behavior of the Tab key is virtually identical to if a normal `<textarea>`/`<input type="text">` were focused (the only exception is that the current cursor position is remembered, because when would you possibly want the cursor to be reset to the beginning? Not even the end, which maybe arguably could make sense).

eluberoff:
Yeah, I think that'd probably be better (tab with selection basically being
identical to tab without)

PR #415, opened 2015-05-29
fix chrome display bug

eluberoff:
could only repro when 'f' is the first character in the desmos calculator,
and only in chrome canary. seems like a non-destructive fix.


eluberoff:
bad fix. good fix coming.

PR #417, opened 2015-06-02
remove bad white-space rule

eluberoff:
breaks sums in FF. discussed here: https://github.com/mathquill/mathquill/commit/32a48e4a4f2cd3ecaedcf36faac5c84fe590acc9#commitcomment-11483930

### DIFF
--- a/src/css/editable.less
+++ b/src/css/editable.less
@@ -1,19 +1,5 @@
 .mq-editable-field {
   .inline-block;
-  .mq-root-block {
-    .inline-block;
-    width: 100%;
-    padding: 2px;
-    .box-sizing(border-box); // supported by IE8+: http://caniuse.com/css3-boxsizing
-    white-space: nowrap;
-    overflow: hidden;
-    vertical-align: middle;
-    // &:after is a fix for
-    // https://bugs.webkit.org/show_bug.cgi?id=35443#c1
-    &:after {
-      content: '';
-    }
-  }
   .mq-cursor {
     border-left: 1px solid black;
     margin-left: -1px;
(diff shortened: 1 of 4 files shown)


laughinghan:
Woah, removing `white-space` entirely? Doesn't that break pressing space more than once in a row?

eluberoff:
good call. I'll improve this.

eluberoff:
@laughinghan, what would you think of just using &nbsp; for spaces in rendering as a fix there?

`LatexCmds[' '] = LatexCmds.space = bind(VanillaSymbol, '\\ ', '&nbsp;');`

that seems like it's more semantic html. I'll push that to this branch

src/css/editable.less
```diff
@@ -5,7 +5,6 @@
     width: 100%;
     padding: 2px;
     .box-sizing(border-box); // supported by IE8+: http://caniuse.com/css3-boxsizing
-    white-space: pre;
```
laughinghan: `white-space` defaults to `normal` which causes things to wrap:
image
but if we change this line to `white-space: nowrap` instead, with the `&nbsp;` change, I can't think of any further problems.

eluberoff: does this definitely belong just in mq-editable-field? what happens for statically rendered math?

laughinghan: great question, hadn't thought of that

laughinghan:
This looks good but I'd also like to add a visual test case so we can check that this won't happen in Firefox in the future (first attempt didn't work, come back to this)

src/css/editable.less
laughinghan: thanks for pointing this is out---it's exactly why my first attempt a test case for this PR didn't work

PR #422, opened 2015-06-09
Fix chrome & ipad display bugs

### LINKED ISSUE #1: Figure out how to use issue tracking
According to Joel Spolsky, even a one-person, one-file project needs some kind of bug tracking that's not in the programmers head, a text file at the very least. Just now I was thinking "Hm! With this commit I don't think there are any outstanding bugs! ...as far as I can remember." Which is exactly the problem -- I don't remember. Better start learning how to use this.

### LINKED ISSUE #2: Dynamic Cosmetic Issue: Moving cursor in and out of super/subscripts in parens
If you type ((a_n x^n))) in Opera, moving the cursor in and out of the super and subscripted n's causes noticable changing of vertical alignment of outer parens.

Will update when test cases for WebKit and Gecko are found.

eluberoff:
Fix #1: get rid of weird white cover that shows up in chrome 43 in desmos tables.

Before:
screen shot 2015-06-09 at 2 35 35 pm

After:
screen shot 2015-06-09 at 3 09 30 pm

Fix #2:
make sure a second cursor doesn't show up on iPad when using a real textarea. Fix from here:

### DIFF
--- a/src/css/textarea.less
+++ b/src/css/textarea.less
@@ -13,8 +13,13 @@
     position: absolute; // the only way to hide the textarea *and* the
     clip: rect(1em 1em 1em 1em); // blinking insertion point in IE
     resize: none; // hotfix: https://code.google.com/p/chromium/issues/detail?id=355199#c1
-
   }
+
+  //remove system cursor on iPad
+  .mq-textarea textarea:focus {
+    text-indent: -9999em;
+  }
+
   .mq-selectable {
     width: 1px; // don't "stick out" invisibly from a math field,
     height: 1px; // can affect ancestor's .scroll{Width,Height}
--- a/src/css/textarea.less
+++ b/src/css/textarea.less
@@ -10,12 +10,12 @@
 
   .mq-textarea *, .mq-selectable {
     .user-select(text);
-
     position: absolute; // the only way to hide the textarea *and* the
     clip: rect(1em 1em 1em 1em); // blinking insertion point in IE
-
     resize: none; // hotfix: https://code.google.com/p/chromium/issues/detail?id=355199#c1
 
+  }
+  .mq-selectable {
     width: 1px; // don't "stick out" invisibly from a math field,
     height: 1px; // can affect ancestor's .scroll{Width,Height}
   }


jwmerrill:
Isn't Desmos using spans instead of textareas on the iPad? Why do we need a cursor fix there?

src/css/textarea.less
```diff
@@ -10,12 +10,17 @@
...
+  }
+
+  //remove system cursor on iPad
+  .mq-textarea textarea:focus {
+    text-indent: -9999em;
+  }
 
+  .mq-selectable {
     width: 1px; // don't "stick out" invisibly from a math field,
     height: 1px; // can affect ancestor's .scroll{Width,Height}
```
jwmerrill: According to these comments, setting width and height was fixing real bugs. Did we verify that those are no longer necessary for some reason?

eluberoff: I haven't been able to reproduce that. I guess you'd need a mathquill field
that's smaller than 13px by 13px, which at least for Desmos hasn't arisen.
Setting to width or height to anything under (I think) 13px triggered this
chrome display bug for me.

laughinghan: I'm able to repro the following on prod right now:
1. open the Settings popover
2. focus one of the "Step:" math fields
3. scroll rightward
image
(One of the DCG unit tests used to fail with something like the above screenshot; but Jason said you guys completely changed your testing system a while back, to a dedicated Windows VM or something, right? I think it was before that, I'd guess that's why your change didn't cause the same test to fail in the same way. Anyway, we should add a test specifically for this.)

The default size of a textarea's is a lot bigger than 13px by 13px, in Chrome it's 198px by 32px as you can see if you click the "Show Textareas" button in the top right corner of the visual tests page (<http://localhost:9292/test/visual.html>); but it's not the size of the math field that matters, it's the size of the smallest ancestor with `overflow-y` that's not `visible` that will see the mysterious increase in `scrollWidth`, such as in this case, [the Settings popover](https://github.com/desmosinc/knox/blob/893797f42f8ae72550d66fa4f301aae692254e25/frontend/desmos/public/assets/styl/popovers.styl#L26-L41) (by the way, we set `.dcg-popover .dcg-popover-interior` to `overflow: auto` in 2 different places).

laughinghan:
I'm unable to repro the weird white blotches in Chrome, and I'm pretty sure I did modify the relevant CSS:
image

Are you still able to repro? As mentioned above, your fix re-introduces a Settings popover scrolling area bug: https://github.com/mathquill/mathquill/pull/422#discussion_r35135114

For the record, I totally believe you that the 1px by 1px triggered utterly befuddling transient issues in Webkit-descendants, at one point I, too, removed these rules partly because of a bug in Chrome and Safari where the page would [jump around](https://github.com/mathquill/mathquill/pull/261#issuecomment-38333601) when you clicked on some math fields (but not others).

By #361 when we added back the 1px by 1px to fix the Settings popover bug, I couldn't repro the page-jumping-around bug and tried pretend it never happened.

laughinghan:
I think this can be closed. #452 fixes the iPad blinking blue cursor issue; I can't reproduce the white blotches issue, and the proposed fix for it introduces another bug.

(It might even be possible that the white blotches issue, though basically unrelated to the iPad blinking blue cursor issue, is actually _also_ fixed by #452; the problematic line whose removal reportedly fixes the white blotches issue used to cause [another weird issue that was also fixed by #452](https://github.com/mathquill/mathquill/pull/452#commitcomment-12557835).)

PR #423, opened 2015-06-10
Make cmd + arrow behavior consistent in OSX

ctlusto:
- In all other textareas/inputs, cmd + left/right == home/end

### DIFF
--- a/src/services/keystroke.js
+++ b/src/services/keystroke.js
@@ -45,6 +45,7 @@ Node.open(function(_) {
 
     // Ctrl-End -> move all the way to the end of the root block.
     case 'Ctrl-End':
+    case 'Meta-Right': // so cmd + arrow works like other textareas/inputs in OSX
       ctrlr.notify('move').cursor.insAtRightEnd(ctrlr.root);
       break;
 
@@ -57,6 +58,7 @@ Node.open(function(_) {
 
     // Ctrl-Shift-End -> select to the end of the root block.
     case 'Ctrl-Shift-End':
+    case 'Meta-Shift-Right': // OSX
       while (cursor[R] || cursor.parent !== ctrlr.root) {
         ctrlr.selectRight();
       }
@@ -69,11 +71,13 @@ Node.open(function(_) {
 
     // Ctrl-Home -> move to the start of the current block.
     case 'Ctrl-Home':
+    case 'Meta-Left': // OSX
       ctrlr.notify('move').cursor.insAtLeftEnd(ctrlr.root);
       break;
 
     // Shift-Home -> select to the start of the current block.
     case 'Shift-Home':
+    case 'Shift-Meta-Left': // OSX
       while (cursor[L]) {
         ctrlr.selectLeft();
       }
@@ -81,6 +85,7 @@ Node.open(function(_) {
 
     // Ctrl-Shift-Home -> move to the start of the root block.
     case 'Ctrl-Shift-Home':
+    case 'Meta-Shift-Left': // OSX
       while (cursor[L] || cursor.parent !== ctrlr.root) {
         ctrlr.selectLeft();
       }


ctlusto:
In like any other text-typing situation, cmd + arrow moves to beginning/end of line in OSX (like this textarea right here, for instance).  As @jwmerrill pointed out, there's some danger of cmd + arrow becoming forward/back in browser history if you lose focus on the element, but that's always the case, and it doesn't seem to be a big enough problem that any other webapps I use have changed it.

jwmerrill:
This change looks good to me.

My only question is what `meta-left` and `meta-right` usually mean on Linux.

jwmerrill:
@jneen can you weigh in on whether this is also okay/good on Linux?

wolfv:
In Linux, Meta  (Alt) + left / right is always browser back/forth regardless of focus in textarea or not.

jwmerrill:
Ok, so we probably don't want to override that behavior.

Is it crazy to start breaking apart these key command handlers by OS?

wolfv:
Only OS X is crazy ... ;) No, I think it makes sense to have different keyboard shortcuts for different OS's.

laughinghan:
It's not crazy...we need that eventually...I think CodeMirror/ACE/Atom all do it...

ctlusto:
Here's [Ace's solution](https://github.com/ajaxorg/ace/blob/3b85ddfa051527f8feafd29a25f0fda9e95fc406/lib/ace/lib/useragent.js#L64), but they hate it...

```javascript
/*
 * I hate doing this, but we need some way to determine if the user is on a Mac
 * The reason is that users have different expectations of their key combinations.
 *
 * Take copy as an example, Mac people expect to use CMD or APPLE + C
 * Windows folks expect to use CTRL + C
 */
```
Problem is `navigator.platform` [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform), and in fact [the spec allows browsers to return an empty string](https://html.spec.whatwg.org/multipage/webappapis.html#dom-navigator-platform).  That whole `navigator` spec is a mess.  E.g., try `navigator.appCodeName`...it must always return `Mozilla`, no matter what the hell you're running.

Is there a version of feature detection here available in the short term?  Check fire a `metaKey` event and check whether `event.altKey` is also true?  A la [this right here](http://jsbin.com/ziyawo/4/edit?js,console)?  Just spitballing.

jwmerrill:
The user agent string usually contains information about what OS you are on, but UA strings are getting pretty crazy these days.

ctlusto:
Codemirror does [something extremely similar](https://github.com/codemirror/CodeMirror/blob/aa8c3682c19cd6e308e3e68304aea3f2cb57f0d8/lib/codemirror.js#L22-L23) (and the word 'kludge' leads me to believe they also don't like it).

laughinghan:
I love your comment @clust, formatting and links galore, guys, can I keep this one?

This is pretty much what I expect from the internet at this point, haha. There is a BestieJS library for determing platform info like OS: http://mths.be/platform , those tend to be reasonably high quality. However, I would really, extra, super duper love if there's any way of telling Mac `Cmd-Left` "Home" apart from Linux `Meta-Left` "Back" that _doesn't_ involve browser sniffing. Alt = Meta on Linux but not on Mac, right? Any way to use that?

ctlusto:
@laughinghan Did you see the JS bin I linked to above?
http://jsbin.com/ziyawo/4/edit?js,console
Returns `false` for me on my mac...ought to return true on Linux, but I have no way of testing that.
Can we trust jQuery to be reliable in smoothing out the `altKey` and `metaKey` compatibility?  If so, maybe something like that will work without having to sniff the OS directly.

laughinghan:
Whoops, closed by GitHub when I deleted the dev branch, but this was awesome, re-opened as #514.

PR #425, opened 2015-06-12
Bubble a "reflow" when an inequality changes to/from strict.

jwmerrill:
Version of https://github.com/mathquill/mathquill/pull/317 target toward dev.


jwmerrill:
Note: only bubbles `reflow` instead of `finalizeInsert`. Is that sufficient?

jwmerrill:
See discussion in #317.

There's a case made in that discussion that "reflow" should be separate from "input/edited", but I think we should have the inequality case trigger a reflow like everything else does until we can come to a more satisfactory alternative to listening to reflow.

Avoiding a reflow in this one particular case is almost certainly not a meaningful performance win.

eluberoff:
agree re: perf, especially since this particular reflow will always be triggered in response to a user action

Issue #429, opened 2015-06-24
[dev] TextBlock's latex changes to \text{undefined} after moving cursor to the left


eoghanmcilwaine:
Hi, I noticed this behaviour on dev. I added [a simple failing test on my fork](https://github.com/Learnosity/mathquill/commit/63c50e8df50269752c2523b6a587db8ad17f2c79) to illustrate.

To test manually:
- Input `\text{x}` (by any method)
- Using arrow key, move cursor back to the left of the `x`
- Latex output is now `\text{undefined}`

PR #430, opened 2015-06-25
Fix TextBlock cursor movement bug

eoghanmcilwaine:
When the cursor is in a certain position, firstChild picks up the .mq-cursor span rather than the text node, causing [issue 429](https://github.com/mathquill/mathquill/issues/429).

My code will still throw an error on line 174 if there's no textNode child. I think that's ok as we already make the assumption elsewhere that TextBlock always has content.

Sorry about the magic number for textNode - maybe there's a better way to filter this?

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -170,7 +170,7 @@ var TextBlock = P(Node, function(_, super_) {
   function fuseChildren(self) {
     self.jQ[0].normalize();
 
-    var textPcDom = self.jQ[0].firstChild;
+    var textPcDom = self.jQ.contents().filter(function (i, el) { return el.nodeType === 3; })[0]
     var textPc = TextPiece(textPcDom.data);
     textPc.jQadd(textPcDom);
 
--- a/test/unit/text.test.js
+++ b/test/unit/text.test.js
@@ -53,4 +53,16 @@ suite('text', function() {
     ctrlr.moveRight();
     assertSplit(cursor.jQ, 'abc', null);
   });
+
+  test('does not change latex as the cursor moves around', function() {
+    var block = fromLatex('\\text{x}');
+    var ctrlr = Controller({ __options: 0 }, block);
+    var cursor = ctrlr.cursor.insAtRightEnd(block);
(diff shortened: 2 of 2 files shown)


kschaefe:
You can use `window.Node.TEXT_NODE` instead of `3`.  I can't find out how far back support for that goes, but I think it's pretty far back as caniuse doesn't even let you search for it.

Another option would be to specify that value in intro.js as a constant usable by all of MathQuill.  Its magical value being defined once with a comment inline and used by name elsewhere for clarity.

laughinghan:
Whoops, closed by GitHub when I deleted the `dev` branch, but this was good work, re-opened as #516.

laughinghan:
This was a great catch, and a really helpful test case! I'm a worried that the patch just pastes over the problem though. `fuseChildren()` appears to expect the the cursor is not in the `TextBlock` even after your patch, it still disowns its children and adopts a new `TextPiece`; a cursor at the left end of the `TextBlock`, like your patch protects against, would now have a pointer to the disowned `TextPiece`.

I think the underlying problem here is that except for this case, `.blur()` satisfies the contract that by the time it's called, all has been said and done, and the cursor has already left/been hidden (e.g. on intentional blur, [the cursor is hidden before `.blur()` is called on its parent](https://github.com/mathquill/mathquill/blob/c103063a62c85127c7eca33fa153a78a3dab6be5/src/services/focusBlur.js#L32)). I implemented a fix for that contract on #528.

About the magic number 3: I actually think it's fine with a comment and citation lik http://reference.sitepoint.com/javascript/Node/nodeType

eoghanmcilwaine:
Ah I see, yeah that looks like a more logical fix! Glad the test case was useful.

I think IE < 9 is the only reason to not use `Node.TEXT_NODE` as @kschaefe suggested. It's cool to pay homage to [ancient times](https://github.com/mathquill/mathquill/blob/7535f2c657c5a5b40774987e9c72e4fe9ebf85ff/src/commands/text.js#L175) anyway :grin:

PR #432, opened 2015-06-29
Fix quadratic jQ.add fold in fragment construction.

jwmerrill:
`jQ.add` sorts the collection it outputs according to document order each time
it is called, so building a collection by folding `jQ.add` directly takes more
than quadratic time in the number of elements.

See the definition of `add` in jQuery 2.1.4

https://github.com/jquery/jquery/blob/2.1.4/src/traversing.js#L112

Here's a jsbin that demonstrates the problem, by timing rendering on
successively larger inputs:

http://jsbin.com/qeseye/edit?js,output

In order for this jsbin to work, you need to build mathquill with `make basic`,
and then serve the build folder on port 8000 (I do this with
`python -m SimpleHTTPServer`, run from the build folder).

Here are the timing results, before and after the change in this commit:

before:
```
   n  time (ms)
   1, 1.8968580000000088
   2, 1.995256999999981
   5, 3.0783539999999903
  10, 5.028018000000003
  20, 8.587542000000013
  50, 22.979758000000032
 100, 60.168756
 200, 189.10946500000006
 500, 965.212992
1000, 3688.4084700000003
2000, 15294.156173
```

after:
```
   n  time (ms)
   1, 3.0877390000000275
   2, 2.8989660000000015
   5, 2.561802
  10, 4.606090999999992
  20, 6.846581999999984
  50, 12.924725999999993
 100, 28.231186999999977
 200, 37.18834900000002
 500, 81.04900400000008
1000, 163.03707799999995
2000, 299.762073
```

And here's a Desmos graph:

https://www.desmos.com/calculator/15ariaibuv

### DIFF
--- a/src/tree.js
+++ b/src/tree.js
@@ -252,7 +252,7 @@ var Fragment = P(function(_) {
     //
     // https://github.com/jquery/jquery/blob/2.1.4/src/traversing.js#L112
     var accum = this.fold([], function (accum, el) {
-      Array.prototype.push.apply(accum, el.jQ.get());
+      accum.push.apply(accum, el.jQ.get());
       return accum;
     });
 
--- a/src/tree.js
+++ b/src/tree.js
@@ -244,7 +244,19 @@ var Fragment = P(function(_) {
     this.ends[dir] = withDir;
     this.ends[-dir] = oppDir;
 
-    this.jQ = this.fold(this.jQ, function(jQ, el) { return jQ.add(el.jQ); });
+    // To build the jquery collection for a fragment, accumulate elements
+    // into an array and then call jQ.add once on the result. jQ.add sorts the
+    // collection according to document order each time it is called, so
+    // building a collection by folding jQ.add directly takes more than
+    // quadratic time in the number of elements.
+    //
+    // https://github.com/jquery/jquery/blob/2.1.4/src/traversing.js#L112
+    var accum = this.fold([], function (accum, el) {
+      Array.prototype.push.apply(accum, el.jQ.get());
+      return accum;
+    });
+
+    this.jQ = this.jQ.add(accum);
   };
   _.jQ = $();
 


jwmerrill:
Graph of rendering time before (orange) and after (blue) this change.

timing graph

src/tree.js
```diff
@@ -244,7 +244,19 @@ var Fragment = P(function(_) {
...
-    this.jQ = this.fold(this.jQ, function(jQ, el) { return jQ.add(el.jQ); });
+    // To build the jquery collection for a fragment, accumulate elements
+    // into an array and then call jQ.add once on the result. jQ.add sorts the
+    // collection according to document order each time it is called, so
+    // building a collection by folding jQ.add directly takes more than
+    // quadratic time in the number of elements.
+    //
+    // https://github.com/jquery/jquery/blob/2.1.4/src/traversing.js#L112
+    var accum = this.fold([], function (accum, el) {
+      Array.prototype.push.apply(accum, el.jQ.get());
```
laughinghan: I feel like just `accum.push.apply(` would make slightly more sense in this case

laughinghan:
This is genius!

laughinghan:
I bet it would be fascinating to hear about how you isolated the perf problem to the `jQ.add()`; I guess this is a variant of the classic quadratic string accumulator problem.

jwmerrill:
It was, perhaps surprisingly, not that hard to track this one down.

Evan Weinberg tweeted a graph at us that used to work, but started to not even load https://twitter.com/emwdx/status/615482918765334528. I just opened up Chrome's dev tools in the hung graph and hit the pause/play button a few times and noticed that every time I paused I was down in some jQuery sorting routine, and a few frames up the call stack was MathQuill's Fragment constructor. From there, like you suggest, it was a fairly obvious pattern match against other quadratic append situations I've found myself in.

I did end up spending a bit of time trying to understand why we didn't run into this same problem before. I looked through a bit of jQuery history, and there have been some changes in how often this sorting is enforced (used to be sometimes, now it's always), but somehow, the release version of mathquill doesn't seem to exhibit the issue, even with jQuery 2.1.4, and I'm not really sure why.

PR #434, opened 2015-07-08
Enforce sub-then-sup order when deleting SupSub blocks

laughinghan:
Say you've got

    x_{sub}^{|sup}

where the | is the cursor. If you hit backspace, what should you get?

    x_{sub}|sup

right?

Well, prior to this commit, you'd get

    x|sup_{sub}

which maybe would make sense if subscripts and superscripts were
semantically on equal footing (e.g. Einstein notation or chemical
symbols symbol), but usually subscripts are part of a variable's name
(e.g. the index), whereas superscripts are exponents.

We already enforce the subscript-then-superscript ordering when
exporting LaTeX and when moving the cursor around with the left and
right arrow keys (when `leftRightIntoCmdGoes` isn't set), so it makes
sense to be consistent with that when deleting.

### DIFF
--- a/test/unit/SupSub.test.js
+++ b/test/unit/SupSub.test.js
@@ -158,10 +158,10 @@ suite('SupSub', function() {
       assert.equal(mq.latex(), 'x_a^b');
 
       mq.keystroke('Left Backspace');
-      assert.equal(mq.latex(), 'xb_a');
+      assert.equal(mq.latex(), 'x_ab');
 
       mq.typedText('c');
-      assert.equal(mq.latex(), 'xcb_a');
+      assert.equal(mq.latex(), 'x_acb');
     });
   });
 });
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -246,12 +246,13 @@ var SupSub = P(MathCommand, function(_, super_) {
     // like 'sub sup'.split(' ').forEach(function(supsub) { ... });
     for (var i = 0; i < 2; i += 1) (function(cmd, supsub, oppositeSupsub, updown) {
       cmd[supsub].deleteOutOf = function(dir, cursor) {
-        cursor.insDirOf(dir, this.parent);
+        cursor.insDirOf((this[dir] ? -dir : dir), this.parent);
         if (!this.isEmpty()) {
-          cursor[-dir] = this.ends[dir];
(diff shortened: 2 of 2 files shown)


laughinghan:
"Prior to this commit" meaning #274, which was an improvement but, c'mon, backspacing "through" the exponent must have felt weird, what was I thinking.

laughinghan:
This fixes the

> * `x_2^3` (in normal, not-auto-subscripting mode) peels off the exponent and then places the cursor
between the base and the subscript. repeated backspacing leaves behind the subscript. a little odd

issue mentioned by @eluberoff in 4a8779f76f9e6c26fc8bb389e9aa5a8f50efe1b7 (#411).

test/unit/SupSub.test.js
```diff
@@ -158,10 +158,10 @@ suite('SupSub', function() {
       assert.equal(mq.latex(), 'x_a^b');
 
       mq.keystroke('Left Backspace');
-      assert.equal(mq.latex(), 'xb_a');
+      assert.equal(mq.latex(), 'x_ab');
```
jwmerrill: This makes sense if you look at the latex directly, but it seems like an odd user experience that something would move from the exponent to the subscript when you press the delete key. It feels like in this case we should maybe move the cursor without changing structure.

laughinghan: it doesn't move to the subscript, we just output LaTeX with all unnecessary spaces removed even if they improve usability; that LaTeX is equivalent to `x_a b` or `x_{a}b`. There's an open issue somewhere to output `x_a b` instead, which I agree with, but I'm not quite sure how

jwmerrill: Sorry, I was misinterpreting what's happening here. `b` isn't being moved into the subscript, it's just being moved to being a sibling of x.

I guess that's fine? I don't really know what a user is trying to accomplish when they delete from the left end of an exponent that has other things in it.

jwmerrill: I'm going to go ahead and merge this because the new behavior seems better than the old behavior. But I think we might want to revisit what happens when you delete from the left side of non-empty blocks in general. E.g.

`1+\frac{|1}{2} -> 1+|12`

also seems weird and not very useful to me.

PR #436, opened 2015-07-09
added npm package files

aaivazis:


laughinghan:
Hi, thanks for your interest in MathQuill!

I actually deliberately added `/build` to the `.gitignore` because I don't want the built JS and CSS tracked by the git repo. That would violate the [Single Source of Truth](https://en.wikipedia.org/wiki/Single_Source_of_Truth) principle: the source files and built files can get out of sync, when someone changes the source but doesn't run `make`, and commits the changes to the source files only, and no one would even notice.

Issue #341, opened 2014-10-15
License explanation is confusing me


hyandell:
Hi,

The bottom of the readme explains the MPL 2.0 by saying "The quick-and-dirty is you can do whatever as long as modifications to MathQuill itself are in public GitHub forks."

That seems to clash with MPL 2.0's FAQ (https://www.mozilla.org/MPL/2.0/FAQ); specifically Q6 which says you can modify privately without having to make a public release, and Q9 and Q10 which don't require modifications be made as GitHub forks.

Could I suggest clarifying this by saying:

 "The quick-and-dirty is you can do whatever as long as distributed modifications to MathQuill itself are made public, for example in GitHub forks."

Thanks.

laughinghan:
Yeah, I said "if" not "iff", what could possibly be confusing about that?

I'll try to clarify by adding a link to the MPL 2.0 FAQ. I don't want to complicate the "quick-and-dirty" explanation, that's targeted at the most common case of people playing around in public forks.

PR #439, opened 2015-07-15
Updated the license summary with a link to tldr legal

robinwassen:
tl;dr legal provides good summaries of what different licenses mean, better to use that than some custom summary.

Solves issue #341


laughinghan:
Good suggestion, although TL;DR Legal is more aimed at maintainers than users. I'ma circle back to this

laughinghan:
Linking to TL;DR Legal doesn't really address either of the clarifications brought up in #341, I'm linking to the MPL 2.0 FAQ instead: 4e1c99d2d72e6801d2ce289286f16d0d63662d98

PR #442, opened 2015-07-20
Only break out of the end of exponents

laughinghan:
...don't jump from the middle of the exponent out to the right of the
exponent.


jwmerrill:
Awesome, good catch.

PR #447, opened 2015-08-03
Polish/fix grouping in ie and ff mathquill lrn 4758

markkellyeire:

PR #450, opened 2015-08-04
Lrn 4726/polyatomic render

mayaijung:


jwmerrill:
https://github.com/mathquill/mathquill/pull/447#issuecomment-127279122

PR #451, opened 2015-08-05
Lrn 4045/florin spacing issue

mayaijung:


jwmerrill:
https://github.com/mathquill/mathquill/pull/447#issuecomment-127279122

PR #452, opened 2015-08-06
Fix blinking blue cursor, autocapitalize on iOS

laughinghan:


jwmerrill:
Awesome. Nice research, as usual.

PR #453, opened 2015-08-06
Feature: incremental backspace

laughinghan:
Continuation of @eluberoff's #411:
> On this branch, backspace into a complex element (fraction, exponent, etc) behaves like a left arrow instead of selecting.
> to me, it's a little more intuitive. A few edge cases feel a little off, but the whole thing feels more intuitive to me.

This new branch:
- merges in `dev` which includes in particular #434, which fixes the odd edge case where you can backspace "through" an exponent if there's also a subscript
- fixes `Ctrl-Backspace`
- attempts to fix @jwmerrill's complaint about deleting empty exponents


kschaefe:
This actually maps to what our old Java-based math editor was doing.  So, this is a big win for us.  Thanks for this.

PR #459, opened 2015-08-18
Interface Versioning

laughinghan:
So I'm really dissatisfied with SemVer and maybe kinda accidentally went down this rabbit hole of coming up with my own versioning system ([InterVer][]) to deal with SemVer's perceived flaws in edge cases that are entirely theoretical for MathQuill...

[InterVer]: https://gist.github.com/laughinghan/009e300626f328792815

All that being said I'm actually really excited about this idea, I'm ready to release current `dev` as v0.10 blocked on just this and docs ([v0.10.0 Milestone][]).

[v0.10.0 Milestone]: https://github.com/mathquill/mathquill/milestones/v0.10.0

### DIFF
--- a/src/outro.js
+++ b/src/outro.js
@@ -1,2 +1,13 @@
+for (var key in MathQuill) (function(key, val) {
+  if (preInterVerMathQuill[key]) return; // already set .noConflict
+  if (typeof val === 'function') {
+    preInterVerMathQuill[key] = function() {
+      insistOnInterVer();
+      return val.apply(this, arguments);
+    };
+    preInterVerMathQuill[key].prototype = val.prototype;
+  }
+  else preInterVerMathQuill = val;
+}(key, MathQuill[key]));
 
 }());
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ for example:
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>
 <script src="/path/to/mathquill.js"></script>
 <script>
+  MathQuill.interfaceVersion(1);
   $('.static-math').each(function() { MathQuill.StaticMath(this); });
   $('.math-field').each(function() { MathQuill.MathField(this); });
 </script>
@@ -38,6 +39,12 @@ To load MathQuill,
 
 [Google CDN-hosted copy]: http://code.google.com/apis/libraries/devguide.html#jquery
 [the latest tarball]: http://mathquill.com/downloads.html
+
+To use the MathQuill API, first declare an interface version:
+
+```js
+MathQuill.interfaceVersion(1);
+```
 
 Now you can call `MathQuill.StaticMath()` or `MathQuill.MathField()`, which
 MathQuill-ify an HTML element and return an API object. If the element had
@@ -65,10 +72,6 @@ HTML element of a static math or math field, returns its API object (if not,
 MathQuill(mathFieldSpan) === mathField // => true
 MathQuill(mathFieldSpan) === MathQuill(mathFieldSpan) // => true
 ```
-
-`MathQuill.noConflict()` resets the global `MathQuill` variable to whatever it
-was before, and returns the `MathQuill` function to be used locally or set to
-some other variable, _a la_ [`jQuery.noConflict()`](http://api.jquery.com/jQuery.noConflict).
 
 Any element that has been MathQuill-ified can be reverted:
 
@@ -120,6 +123,19 @@ Additionally, descendants of `MathQuill.EditableField` (currently only
 [on `textarea`s]: http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-48880622
 [on `input`s]: http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-34677168
 [key values]: http://www.w3.org/TR/2012/WD-DOM-Level-3-Events-20120614/#fixed-virtual-key-codes
+
+MathQuill overwrites the global `MathQuill` variable when loaded. You can undo
+that with `.noConflict()` (similar to [`jQuery.noConflict()`]
+(http://api.jquery.com/jQuery.noConflict)):
+
+```html
+<script src="/path/to/first-mathquill.js"></script>
+<script src="/path/to/second-mathquill.js"></script>
+<script>
+var secondMathQuill = MathQuill.interfaceVersion(1).noConflict();
+secondMathQuill.StaticMath(...);
+</script>
+```
 
 #### Configuration Options
 
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -23,8 +23,6 @@ MathQuill.noConflict = function() {
   window.MathQuill = origMathQuill;
   return MathQuill;
 };
-var origMathQuill = window.MathQuill;
-window.MathQuill = MathQuill;
 
 /**
  * Returns function (to be publicly exported) that MathQuill-ifies an HTML
@@ -179,3 +177,38 @@ function RootBlockMixin(_) {
     _[name] = function(dir) { this.controller.handle(name, dir); };
   }(names[i]));
 }
+
+/**
+ * Interface Versioning (#459) to allow us to virtually guarantee backcompat.
+ * v0.10.x introduces it, so for now, don't completely break the API before
+ * MathQuill.interfaceVersion(1) is called, just complain with console.warn().
+ *
+ * .noConflict() is shimmed here directly because it needs to be modified,
+ * the rest are shimmed in outro.js so that MathQuill.MathField.prototype etc
+ * can be accessed (same reason this is at the end of publicapi.js, so that
+ * MathQuill.prototype can be accessed).
+ */
+function insistOnInterVer() {
+  if (window.console) console.warn(
+    'Please call MathQuill.interfaceVersion(1) before doing anything else ' +
+    'with the MathQuill API. This will be required starting v1.0.0.'
+  );
+}
+function preInterVerMathQuill(el) {
+  insistOnInterVer();
+  return MathQuill(el);
+};
+preInterVerMathQuill.prototype = MathQuill.prototype;
+
+preInterVerMathQuill.interfaceVersion = function(v) {
+  if (v !== 1) throw 'Only interface version 1 supported. You specified: ' + v;
+  return window.MathQuill = MathQuill;
+};
+
+preInterVerMathQuill.noConflict = function() {
+  insistOnInterVer();
+  window.MathQuill = origMathQuill;
+  return preInterVerMathQuill;
+};
+var origMathQuill = window.MathQuill;
+window.MathQuill = preInterVerMathQuill;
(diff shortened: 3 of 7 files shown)


jwmerrill:
Just read the InterVer document, and looked at the code changes here.

It seems like an interesting proposal, but I don't feel super qualified to make a call about whether it's a "good idea" or not.

Could you write a bit of narrative describing some hypothetical future changes to Mathquill where this feature will benefit users, and what they should do when?

laughinghan:
Great point, I should've talked about this. I dunno about hypothetical future changes, but I do have in mind a concrete use for this.

I've [mentioned](https://github.com/mathquill/mathquill/pull/317#issuecomment-122919297) an intention for event handlers to be passed an equivalent of QuillJS' [`source`](http://quilljs.com/docs/events/) (no relation), but it's awkward because our current signature for event handlers already has an optional argument, the math field itself that's emitting the event (the `MathList` example in README). Instead, I'd rather pass an event object to event handlers, on which there can be any number of properties summarily ignored by event handlers. Plus there'd be no weird non-uniform `handler(mathField)` vs `handler(dir, mathField)` signatures.

But these event handlers have been on `dev` for over a year, I can't just change their signature willy-nilly. With this system in place I can have interface version 2 change the signature, while interface version 1 continues to use the current signature, so people can upgrade to the latest MathQuill without their event handlers breaking.

There's also a concrete historical API change that I wish used this, #336, which systematized the bad old _ad hoc_ global config methods:
> - [...] e.g. `MathQuill.addAutoCommands('sin cos log')` is now `MathQuill.config({ autoCommands: 'sin cos log' })`

Right from the beginning we all fully expected to move eventually to a proper config architecture. If InterVer had been in place, we could trust that when the config architecture came along, we could bump the interface version and continue to support the _ad hoc_ methods under the legacy interface version. Instead it was one more reason to drag my feet about merging `dev` into `master`&mdash;if we get the config architecture in soon enough, we can pretend the _ad hoc_ methods never happened!

src/outro.js
```diff
@@ -1,2 +1,13 @@
+for (var key in MathQuill) (function(key, val) {
+  if (preInterVerMathQuill[key]) return; // already set .noConflict
+  if (typeof val === 'function') {
+    preInterVerMathQuill[key] = function() {
+      insistOnInterVer();
```
jwmerrill: I'm a little confused about how this will work out in practice. Does this mean that people who don't specify an interfaceVersion will see a console warning for each API call they make?

Is it considered valid to not declare an interface version until you are ready to opt in to version 1? Or is version 1 effectively the same as the pre InterVer behavior, and we're expecting that everyone that gets on "MathQuill 2.0" (if that's what we're calling the dev branch once it's released) should definitely immediately specify an InterVer version.

In any case, I think it would make sense to reduce the noise a little bit, and only do this warning either once, or once per different API method call.

jwmerrill:
Let me try to summarize what InterVer buys you:

With both SemVer and InterVer, MAJOR version bumps are pretty much "anything goes." You can (not should, of course...) break backward compatibility any way you like, as long as you document it, and a user should generally expect to do some work when they upgrade MAJOR versions.

The issue is that this is often, in practice, actually *too much* flexibility. Developers are generally very hesitant to bump MAJOR version numbers, because this signals to their users that they might have to do a bunch of work to upgrade. This leads to either stagnation (breaking changes that should happen don't because it's too much trouble), violations of SemVer by not bumping the MAJOR version for every breaking change (but this is a *small* breaking change...), or buildups of risk where a MAJOR version bump branch is created and developed, but it takes a long time to release and adopt because it's all-or-nothing. SemVer is especially likely to be violated by "fixing a bug" in the interface that developers were actually relying, and only incrementing the MINOR version number.

InterVer gives you a structured way to allow users to upgrade versions that allow *opting in* to an interface change, but don't immediately require it. This lets users benefit from non interface-changing improvements, and then make the interface change as they find it convenient.

Another way to achieve the same goal is to backport changes that don't require interface changes from the latest MAJOR release to earlier MAJOR releases. But running that process well is tedious and expensive in terms of developer time.

jwmerrill:
When is it appropriate to deprecate interface version numbers? Do you have to do a MAJOR version increment in order to drop old interface versions?

src/outro.js
laughinghan: > I'm a little confused about how this will work out in practice. Does this mean that people who don't specify an interfaceVersion will see a console warning for each API call they make?

Yes.

> Is it considered valid to not declare an interface version until you are ready to opt in to version 1? Or is version 1 effectively the same as the pre InterVer behavior, and we're expecting that everyone that gets on "MathQuill 2.0" (if that's what we're calling the dev branch once it's released) should definitely immediately specify an InterVer version.

Declaring interface version 1 changes no API behavior except to stop the warnings.

The idea is basically that the easiest thing to do with an API should be the thing that the API's designers want the vast majority of people to do. Things intended for a minority shouldn't be the path of least resistance. That's why declaring an interface version has to be required, we can't just have omitting an interface version be equivalent to declaring the pre-InterVer interface version: the pre-InterVer interface version would be the path of least resistance.

All you have to do to stop the warnings is add that one `MathQuill.interfaceVersion(1);` call. Given that, do you feel strongly about reducing the warnings?

laughinghan:
> When is it appropriate to deprecate interface version numbers? Do you have to do a MAJOR version increment in order to drop old interface versions?

Yes.

> With both SemVer and InterVer, MAJOR version bumps are pretty much "anything goes." You can break backward compatibility any way you like, as long as you document it, and a user should generally expect to do some work when they upgrade MAJOR versions.

Well, not quite. With InterVer, even MAJOR version bumps still aren't "anything goes", breaking changes for a given interface version are still verboten (except when they'd be allowed for a minor version bump too, like a security fix; of course, the same fix should also be made to affected interface versions of the previous major version (in a new minor version)).

The only breaking change that MAJOR version bumps are allowed is in the SemVer sense, which is to drop support for some old interface versions. Other than dropping old interface versions (and exceptional cases like security fixes), InterVer doesn't allow any breaking changes in the SemVer sense ever, between minor OR major version bumps. Code written against a declared interface version should work with all future releases, or fail very early complaining about a dropped interface version.

> Developers are generally very hesitant to bump MAJOR version numbers, because this signals to their users that they might have to do a bunch of work to upgrade.

Well, with InterVer, users on the latest interface version shouldn't have to do any work at all to upgrade, and hopefully many are only a little behind the latest interface version and so would only have to do a little work to upgrade. This hope could be wrong: it's possible that in reality the vast majority of users are on the earliest interface version, and so typically will have to do a bunch of work to upgrade. This seems unlikely to me though if minor versions routinely introduce useful new features with small incompatibilities, that users routinely do a little interface version upgrade to take advantage of. (For example, the thing where you could tell whether an `edited` event was from the typist or the API? That works best if we switch event handlers to taking in an event object rather than many parameters. That's exactly the kind of thing I plan for interface version 2.)

> This leads to [...] buildups of risk where a MAJOR version bump branch is created and developed, but it takes a long time to release and adopt because it's all-or-nothing.

Exactly. Don't forget the Python 3 scenario, where the bumped MAJOR version is actually released on a reasonable schedule, but because upgrading to it is basically all-or-nothing, it doesn't see widespread adoption even years later. (It's not literally all or nothing, there are 5 Python 3 incompatibilities you can [import `from __future__`](https://docs.python.org/2/library/__future__.html); but I'd hazard that there's a lot more than 5 incompatibilities between Python 2 and 3.)

> InterVer [...] lets users benefit from non interface-changing improvements, and then make the interface change as they find it convenient.

_And_ do so piecemeal rather than all-or-nothing.

> Another way to achieve the same goal is to backport changes that don't require interface changes from the latest MAJOR release to earlier MAJOR releases. But running that process well is tedious and expensive in terms of developer time.

Hmmmm...I suppose we could actually consider the MAJOR version number to _be_ the interface version, and whenever we release patches, we do so to several MAJOR versions at once. With a sufficiently confusing build system this might not even be that tedious, we could have "compiler" directives to mark certain sections of code as being targeted towards certain ranges of MAJOR/interface versions, and whenever we release patches, we compile the current version of `master` to several different MAJOR versions at once, which differ only as directed in those certain sections.

Not only would that be bizarre and difficult to explain, we want upgrading interface versions to be a very low-resistance path, like, changing one number passed to a MathQuill API function and not having to touch the dependencies system at all (Bower/NPM/Git Submodules).

jwmerrill:
Let's try it!

kschaefe:
Does the interface versioning only cover method signatures or does it also cover configuration property changes?  Our change in https://github.com/apollo-ctp/mathquill/pull/11 overloads spacesBehaveLikeTabs to allow a string property as well.  It would be better if the entire option were renamed to spaceBehavior to allow a more dynamic range of options, such as our new approach therein.

laughinghan:
Also configuration property changes. The theory is you should be able to upgrade your MathQuill version and not change a single line of code and everything should keep working as before.

(_Tangent_: One wrinkle in this is `.keystroke()` and `.typedText()`. If not for them, changes like whether Spacebar behaves like Tab, or "Incremental backspace", would only affect typist interaction and never the API. `.keystroke()` and `.typedText()` are equivalent to the typist typing things, which is exactly the right thing when they're used by a virtual keyboard, but also means that if the cursor is after a square root and you're expecting `.keystroke('Backspace Backspace')` to delete the whole square root, well, the behavior of that code has changed.

What I'm currently imagining is that that's part of their contract: don't use them to manipulate math, use `.latex()` and `.write()` and `.cmd()` to manipulate math, `.keystroke()` and `.typedText()` are only for a virtual keyboard, and just like the real keyboard, the typist's interface should improve over time without the API consumer having to bump their interface version (and fix any use of backcompat-breaking APIs).

The other option is 

_Anyways_, if you change what any of `spaceBehavesLikeTab: true`, or `spaceBehavesLikeTab: false`, or leaving it unset does, yes, that change should only be made in a bumped interface version. Which, by design, preserves backcompat but is still pretty easy to do, because we want it to be easy to make small improvements like deprecating `spaceBehavesLikeTab` for `spaceBehavior`.

src/outro.js
kschaefe: This is failing in Mac Chrome 45.0.2454.85 with:
TypeError: Cannot set property 'prototype' of undefined(anonymous function)

The reason for this is `typeof preInterVerMathQuill === 'number'`.  While Chrome does not throw an error, when you assign a property to a number, you cannot actually access that property; it will always be undefined.

The following a Chrome console I tested, each set is my input plus Chrome's output:
```javascript
var test = 1;
undefined

test["special"] = 'something awesome'
"something awesome"

test["special"]
undefined
```

Whatever you're using to store the interface version needs to be a real Object.

src/outro.js
```diff
@@ -1,2 +1,13 @@
...
+      return val.apply(this, arguments);
+    };
+    preInterVerMathQuill[key].prototype = val.prototype;
+  }
+  else preInterVerMathQuill = val;
```
kschaefe: This line is in error, see my comment above.  It should be:
```javascript
else preInterVerMathQuill[key] = val;
```

We will be updating our MathQuill version ASAP.  I will try to get a pull request soon.

PR #462, opened 2015-09-03
`edit`, `edited` as aliases for `reflow` event

laughinghan:
What is currently on the `dev` branch known as the `reflow` event was
called `edited` when it was introduced in #241. @alopatin then brought
up a serious issue (#317) with the `edited` event as it was then
implemented, which is that you could edit < into ≤ (etc) without
triggering the event.

Rather than fixing the problem, I obstinately renamed the event from
`edited` to `reflow` so that I could claim that the event not being
triggered when no sizes change was a feature, not a bug (#325).

@jwmerrill finally cut through my bullshit and got @alopatin's fix
from #317 merged into `dev` (tweaked as #425), which made edits like
< into ≤ call the `reflow` handler, so that there'd be at least some
handler that's reliably called when the contents of a field changes.

Because the `reflow` event was so broken for so long, people like
@alopatin have been using a fork of `dev` with a fixed `edited`.
To ease their migration to v0.10.0, and also because the name makes
more sense for its current role, I want to add back `edited` as an
alias for `reflow` (without removing `reflow`).

While we're talking about a name that makes sense, what name for this
event do we want to recommend going forward? How about we get rid of
the inexplicable past tense and just call the event `edit`. Make that
the recommended alias in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,17 +15,26 @@ Just load MathQuill and call our constructors on some HTML element DOM objects,
 for example:
 
 ```html
-<p>
-  Solve <span class="static-math">ax^2+bx+c=0</span>:
-  <span class="math-field">x=</span>
-</p>
 <link rel="stylesheet" href="/path/to/mathquill.css"/>
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>
 <script src="/path/to/mathquill.js"></script>
 <script>
-  $('.static-math').each(function() { MathQuill.StaticMath(this); });
-  $('.math-field').each(function() { MathQuill.MathField(this); });
+  $(function() {
+    MathQuill.StaticMath($('#problem')[0]);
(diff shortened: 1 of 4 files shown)


eluberoff:
One possibility is "input," which, as I understand it, is supposed to fire
anytime an input's contents change in response to any type of user action.

I guess that's not a perfect synonym for reflow, but would be nicely
parallel to other input fields in modern browsers.

laughinghan:
Yeah but we actually want the event to trigger also if the contents change in response to an API call (e.g. virtual keyboard), whereas `input` doesn't fire if you use the DOM interface to manipulate an element's contents.

Really `change` would be the perfect name if the DOM hadn't already fucked it. We could go the ReactJS route and call it `change` even though it's totally different from the DOM event (as-you-type rather than only on `blur`).

jwmerrill:
I like `edit`.

jwmerrill:
Awesome. Glad to be moving forward with this one.

PR #468, opened 2015-09-23
Use WOFF and WOFF2 font formats.

EvilOatmeal:
The formats are widely supported in browsers today (IE 9+), and the font-face syntax allows for fallbacks where they are not supported.

The WOFF2 files were created by converting the WOFF files using the tool available at https://everythingfonts.com/woff-to-woff2


laughinghan:
Both WOFF 1 and 2? Also, any idea why WOFF1 wasn't used in the first place even though the font file was checked in? I didn't say very much in the commit where I removed it: c9be44bac7bc7a3dab41e728a1efa7c1e2929b7a

EvilOatmeal:
WOFF 1 as a fallback for those browsers that don't support WOFF 2. Only the first supported format will be loaded. Take a look at http://beta.caniuse.com/#search=woff

No idea why WOFF would cause a syntax error in Opera back in 2010. Never heard of it, and I've read most of the popular cross browser web font articles. But Opera uses Blink now, same as Chrome, so this should be a non-issue.

The only Opera issue I know of is "Opera will fail if you do not use quotes (single or double) in your local() font name" as found in [Bulletproof @font-face Syntax by Paul Irish](http://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/), but in the commit you referenced you do use quotes.

laughinghan:
Alright, well, I guess I'll (or anyone can :smile:) do some cross-platform testing, post some screenshots, and if all's well I'll be happy to merge.

robinwassen:
For reference, browser compatibility for WOFF and WOFF2:

[Can I Use - WOFF](http://caniuse.com/#feat=woff)
[Can I Use - WOFF2](http://caniuse.com/#search=woff2)

No compatibility tables for mathquill exists - so I do not know exactly what I am supposed to test with. I will post some tests and screenshots from random major browsers.

robinwassen:
Browser          | Font   | Result   | Screenshot       |
-----------------|--------|----------|------------------|
Chrome 47 (Mac)  | WOFF2  | OK       | chrome-47-woff2 |
Safari 9 (Mac)   | WOFF   | OK       | desktop-safari-9-woff |
Firefox 41 (Mac) | WOFF   | OK       | firefox-41-woff2 |

Looks pretty good to me, I think this is safe to merge.

robinwassen:
@laughinghan Do you have any more browsers in mind that you want compatibility tests for?

laughinghan:
Oh wow, thanks! The Firefox 41 one looks wrong though hopefully it just loaded wrong?

Also, we do need at least one Windows and IE screenshot, I've had a problem where the font was super aliased on Windows (both IE and Chrome, at the time) 6f028517edfa1242e8f41ee2fe19d4969cd94141

robinwassen:
@laughinghan I made a mistake when doing the Firefox Mac OS X tests, used the file:/// protocol which made it not load the font. Retried now over HTTP and it loaded the font as it should, so I replaced the screenshot.

And here are the Windows 8 results:

Browser         			| Font   | Result   | Screenshot       |
----------------------------|--------|----------|------------------|
Chrome 46 (Win 8)  			| WOFF2  | OK       | chrome-46-win-woff2  |
Internet Explorer (Win 8)  	| WOFF   | OK       | ie11-woff  |

I know as you say that custom font rendering were rather horrible on Chrome running on Windows, but it seems like they have improved it quite a lot.

laughinghan:
You're fantastic!

PR #472, opened 2015-09-25
Remove a redundant bug fix for webkit.

### LINKED ISSUE #470: [dev] Regression to fix "Keeps blocks from collapsing to zero width/height when empty by inserting invisible text."
Regression introduced in commit ee4c0b3a742a15da270f6cf51a70aa701f774be9.

The code moved into `math.less` will override the fix in `editable.less` because the rules have the same specificity and the rules have now changed order in the compiled code.

mathquill-height-focus-blur-regression

@jwmerrill, have you found a fix for Desmos without breaking any other stuff?

EvilOatmeal:
Since we're now using "white-space: nowrap" instead of "pre" the bug no longer occurs.

The fix caused a regression after commit ee4c0b3a742a15da270f6cf51a70aa701f774be9.

The regression was that empty MathField's would change their height when focused/blurred.

Closes #470

### DIFF
--- a/src/css/math.less
+++ b/src/css/math.less
@@ -6,11 +6,6 @@
   white-space: nowrap;
   overflow: hidden;
   vertical-align: middle;
-  // &:after is a fix for
-  // https://bugs.webkit.org/show_bug.cgi?id=35443#c1
-  &:after {
-    content: '';
-  }
 }
 .mq-math-mode {
   font-variant: normal;
--- a/test/unit/css.test.js
+++ b/test/unit/css.test.js
@@ -16,7 +16,7 @@ suite('CSS', function() {
     $(mock).css({
       fontSize: '',
       height: '',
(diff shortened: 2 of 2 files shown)


laughinghan:
Oh, really? Sweet!

You were able to repro the bug, and confirm that it doesn't apply with `white-space: nowrap`?

Also, can we add some kind of test case for the empty math field height regression? The unit tests have a stylesheet, could we check with jQuery that the height >0 or something?

EvilOatmeal:
Yes. I will look into adding test cases.

jwmerrill:
Confirmed that this fixes one of Desmos's blockers for getting up to the latest version of dev (broken layout in the settings menu).

Thanks a lot for tracking this down!

jwmerrill:
Tests look good to me. Thumbs up for merge.

Issue #469, opened 2015-09-23
[dev] Deleting inserted broken LaTeX command crashes the editor


robinwassen:
A regression has appeared sometime between July 8 and now that causes the editor to crash if you enter a broken LaTeX command into a MathField.

**Steps to reproduce:**
1. Initialize a MathField using jQuery 1.11.0 and latest dev
2. Focus the MathField and input \brokencmd and press enter
3. Press backspace to delete the command

**Outcome:**

The following error is written to the console:

```
mathquill.js:648 Uncaught TypeError: el.focus is not a function
```

After that the MathField is totally broken and you cannot interact with it, generating errors such as:

```
mathquill.js:1600 Uncaught TypeError: cursor.hide(...).parent.blur is not a function
mathquill.js:614 Uncaught TypeError: this.parent.focus is not a function_.show @ mathquill.js:614(anonymous function) @ mathquill.js:1584n.event.dispatch @ jquery.min.js:3r.handle @ jquery.min.js:3
mathquill.js:614 Uncaught TypeError: this.parent.focus is not a function_.show @ mathquill.js:614_.seek @ mathquill.js:2159(anonymous function) @ mathquill.js:2132n.event.dispatch @ jquery.min.js:3r.handle @ jquery.min.js:3
mathquill.js:1600 Uncaught TypeError: cursor.hide(...).parent.blur is not a function
```

**The test page I used:**

```
<html>
<head>
	<link rel="stylesheet" href="./build/mathquill.css"/>
	<script src="./build/jquery.min.js"></script>
	<script src="./build/mathquill.js"></script>
	<script>
		var mq, f;
		$(function() {
			f = $('.math-field');
			mq = MathQuill.MathField(f[0]);
			mq.write("f(x) =");
		});
	</script>
</head>
<body>
	<div class="math-field" style="padding: 10px; border: 1px solid #000"></div>
</body>
</html>
```

robinwassen:
Did some closer research and found that it is deleting broken LaTeX commands that actually causes the problem, so I would guess that this regression was introduced in the changes August the 6th.

robinwassen:
The reason why I know that it was introduced recently is that I am unable to reproduce this issue when I use the code base at commit: 4fb72dab94163291035432c426593abeb79a0a37

PR #473, opened 2015-09-25
Reverted call to moveTowards and added test case

FWennerdahl:
Fix for #469

The change made in 6648709 caused backspacing into text blocks to break. The change worked fine for MathCommand but TextBlock has another implementation of moveTowards (text.js:r.79).

Included test case for this scenario.

### DIFF
--- a/test/unit/backspace.test.js
+++ b/test/unit/backspace.test.js
@@ -197,6 +197,16 @@ suite('backspace', function() {
     assert.equal(mq.latex(),'n=1');
   });
 
+  test('backspace inside text block', function() {
+    mq.latex('\\text{x}');
+
+    mq.keystroke('Backspace');
+
+    var textBlock = rootBlock.ends[R];
+    assert.equal(cursor.parent, textBlock, 'cursor is in text block');
+    assert.equal(cursor[R], 0, 'cursor is at the end of text block');
+  });
+
   suite('empties', function() {
     test('backspace empty exponent', function() {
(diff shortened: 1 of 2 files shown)


test/unit/backspace.test.js
```diff
@@ -197,6 +197,16 @@ suite('backspace', function() {
...
   });
 
+  test('backspace inside text block', function() {
+    mq.latex('\\text{x}');
+
+    mq.keystroke('Backspace');
+
+    var textBlock = rootBlock.ends[R];
+    assert.equal(cursor.parent, textBlock, 'cursor is in text block');
+    assert.equal(cursor[R], 0, 'cursor is at the end of text block');
```
laughinghan: sweeeeet wanna also assert like, `cursor[L].ctrlSeq` is `'x'` or something?

laughinghan:
You bisected it! This is absolutely awesome!

laughinghan:
@FWennerdahl For future reference if your commit message said "Fix #469" instead of "Fix for #469", GitHub would auto-close that issue and link to the commit in one line item of that issue when I merge this (well...it'll do that once I get rid of the `dev` branch and just have a `master` branch).

FWennerdahl:
Thanks @laughinghan, will do!

One thing though, when running the updated test it fails. Seems like it should be `cursor[L].text` rather than `cursor[L].ctrlSeq`, correct?

laughinghan:
Ugh, yeah.

Issue #474, opened 2015-09-28
[dev] Backspacing out of empty text block makes latex \text{undefined}


FWennerdahl:
Following the revert in PR #473, another bug was discovered. When having the cursor in an empty text block and pressing backspace, the latex of the block becomes `\text{undefined}`. Pressing backspace again seems to remove the block from the DOM but leave the actual block object, still yielding the same latex.

This has been confirmed to have been present at a7eef27.

Working on a fix for this now. The first step was to introduce an isEmpty implementation for TextBlocks. My initial implemenation of this looks as follows:

    _.isEmpty = function() {
        return (this.ends[L] === 0 || this.ends[L].text === undefined) &&
            (this.ends[R] === 0 || this.ends[R].text === undefined);
    };

This allows deletion of the text block but still leaves the issue of giving the `\text{undefined}` latex for empty blocks.

One way to handle this would be to ensure that TextPiece.deleteTowards always leave at least one text block with text set to an empty string. However, the comments in the code mention a requirement that text contents can never be empty.

text.js:r.52-54
```
// the latex \text{} ?  This violates the requirement that
// the text contents are always nonempty.  Should we just
// disown the parent node instead?
```
r.194
`* Text contents must always be nonempty.`

I am unsure what this requirement applies to and where it comes from. Does it mean that TextBlocks always have to contain at least one TextPiece, or that the text of the pieces can never be empty strings? From what I can tell this command comes from the AMSLaTeX system but I could not find any details about this requirement there.

I'll produce test cases to display this behaviour. I'll also work on a solution with an empty text piece until I know more about this requirement as well as investigate how this relates to #313, #126 and #429.

FWennerdahl:
The issue shares root cause with #429. The cursor is found in TextBlock.fuseChildren which does not have a property `data`, resulting in `undefined`being inserted in the TextPiece.

I applied PR #430 which prevents the cursor to be used in fuseChildren, however it did not fully solve the problem for empty text blocks. If no text node is present we instead run into an issue similar to #313.

[...]

PR #476, opened 2015-09-29
Handling of empty TextBlocks

### LINKED ISSUE #126: On dev, leaving an empty \text{} rendered from LaTeX will error
In `\text{NaN}` test case (on `test/visual.html`), backspace or delete forward until empty, then move out with arrow keys or backspace and delete again.

(Same root cause as #121.)

### LINKED ISSUE #313: [dev] Pasting `\text{}` causes an error
Pasting `\text{foo}` works fine, but `\text{}`causes an error in `fuseChildren`: "Cannot read property 'data' of null".

On a related note, typing `\text` and then anything will also cause a JS error (sometimes in the same place as above, sometimes elsewhere).

Found this while fuzz testing (aka key mashing) - some combination of valid key presses seems to insert `\text{}`, but I couldn't track down which one.

### LINKED ISSUE #429: [dev] TextBlock's latex changes to \text{undefined} after moving cursor to the left

FWennerdahl:
Work in progress for #474.

- Fixes #126
- Fixes #313
- Incorporates PR #430 which fixes #429

Outstanding issue with stepping out and back into an empty text block, see failing test added in e433563.

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -48,10 +48,8 @@ var TextBlock = P(Node, function(_, super_) {
     return optWhitespace
       .then(string('{')).then(regex(/^[^}]*/)).skip(string('}'))
       .map(function(text) {
-        // TODO: is this the correct behavior when parsing
-        // the latex \text{} ?  This violates the requirement that
-        // the text contents are always nonempty.  Should we just
-        // disown the parent node instead?
+        if (text.length === 0) return;
+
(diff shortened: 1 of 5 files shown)


FWennerdahl:
Fix for failing test in e433563 applied.

laughinghan:
Whoops, closed by GitHub when I deleted the `dev` branch, but this was good work, re-opened as #517.

PR #479, opened 2015-10-01
Text selection fix

### LINKED ISSUE #477: [dev] Selection in TextBlocks throws errors
When clicking inside a TextBlock and dragging the mouse to select text contents, the following error is thrown:

```
Uncaught TypeError: Cannot use 'in' operator to search for '162' in undefined
_.select @ mathquill.js:751
docmousemove @ mathquill.js:2102
jQuery.event.dispatch @ jquery.js:4435
jQuery.event.add.elemData.handle @ jquery.js:4121
```

This error repeats each time the mousemove event is fired for as long as the mouse button is held. Starting the selection outside the TextBlock works fine, but not when clicking and dragging inside it.

The reason is that `TextBlock.seek` recreates the `Cursor.anticursor` without the `ancestors` map that is created in `Cursor.startSelection`.

I fixed this by moving the creation of the ancestor map to `Point.init` which ensures that it is always available.

This revealed another bug: making a selection in a TextBlock and then clicking somewhere else in the block to make a new selection causes the new selection to start at the same point as the first one. This is because the `Cursor.anticursor` is never not deleted which in turn means the `TextBlock.anticursorPosition` is not reset.

This was resolved by simply calling `Cursor.endSelection` in the mousedown event handler.

I'll try to create some describing tests for these issues and make a pull request with the fixes.

FWennerdahl:
Fixes #477 (which also explains these fixes)

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -147,7 +147,7 @@ var TextBlock = P(Node, function(_, super_) {
       // mouse-selecting within this TextBlock, re-insert the anticursor
       var cursorPosition = cursor[L] && cursor[L].text.length;;
       if (this.anticursorPosition === cursorPosition) {
-        cursor.anticursor = Point.copy(cursor);
+        cursor.anticursor = Anticursor(cursor.parent, cursor[L], cursor[R]);
       }
       else {
         if (this.anticursorPosition < cursorPosition) {
@@ -157,7 +157,7 @@ var TextBlock = P(Node, function(_, super_) {
         else {
           var newTextPc = cursor[R].splitRight(this.anticursorPosition - cursorPosition);
         }
-        cursor.anticursor = Point(this, newTextPc[L], newTextPc);
+        cursor.anticursor = Anticursor(this, newTextPc[L], newTextPc);
       }
     }
   };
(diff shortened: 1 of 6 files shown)


laughinghan:
Thanks for all the great work on `TextBlock`s! Thanks for the tests, I wasn't sure how to test mouse events.

I have some concerns about the specifics of these particular fixes. By moving the ancestor map to `Point::init` you're adding it to the cursor too, where it's not updated and quickly becomes out-of-date. Also, I think it's pretty weird to call `cursor.endSelection()` and then immediately `cursor.startSelection()` in `mousedown`.

Here's some suggested fixes: https://github.com/FWennerdahl/mathquill/pull/3

FWennerdahl:
Thanks for the feedback and the updates. I tested and looked through the changes and they're all good. The Anticursor class is a much nicer solution and I agree that the tests shouldn't be concerned with underlying implementation details such as how the anticursor is updated.

Nothing more to add from my part regarding this!

laughinghan:
Whoops, closed by GitHub when I deleted the dev branch, but I still think this was a good idea, re-opened as #512.

PR #484, opened 2015-10-13
Fix broken links, unlinked links in README

aegamesi:


laughinghan:
Oof, good catches, thanks!

PR #487, opened 2015-10-19
Enabled autoscroll for API-calls cmd and write

FWennerdahl:
When calling public API methods cmd and write the automatic horizontal scroll is not enabled, causing overflowing input to remain hidden for editors that have a defined max width.

Calling `controller.scrollHoriz()`after input has been added solves this.


laughinghan:
Woah, really good point. Can we also add tests for this?

FWennerdahl:
Tests added.

robinwassen:
@laughinghan - What is the status on merging this?

laughinghan:
Whoops, closed by GitHub when I deleted the `dev` branch, but this was really good work, re-opened as #519.

PR #495, opened 2015-12-23
Feature: MathQuill.getInterface()

### LINKED ISSUE #464: Can't specify interface version more than once
Imagine that you want to use two third-party scripts that uses MathQuill on the same page.

They both have to define the interface version they want to use, because they're third-party and don't know about the other scripts on the page.

But currently you can't specify the interface version more than once.

When the second third party script tries to specify the interface version the method to do so is not available.

```js
function first3rdParty(el, latex) {
	var mq = MathQuill.interfaceVersion(1);
	mq.MathField(el).latex(latex);
}

function second3rdParty(el, latex) {
	var mq = MathQuill.interfaceVersion(1);
	mq.MathField(el).latex(latex);
}
```

laughinghan:
Fix #464: as [pointed out](https://github.com/mathquill/mathquill/issues/464#issuecomment-166861641), the `MathQuill.interfaceVersion()` declaration results in a version dependency hell if an app uses third-party scripts that also use MathQuill. Now, no more mutation, instead `MathQuill.getInterface()` nonmutatively creates a whole new copy of the entire API,
closing over the specified interface version:
```js
var MQ1 = MathQuill.getInterface(1);
var MQ2 = MathQuill.getInterface(2);
// MQ1 and MQ2 are two wholly separate instances of the entire API,
// with different interface versions

var mq1 = MQ1.MathField(mathFieldSpan);
var mq2 = MQ2(mathFieldSpan);
// mq1 and mq2 are API objects for the same math field, but they
// have different interface versions
```

There were 2 main sacrifices:
- There's no longer a unique API object per actual, "physical" instance of a math field or static math, `mq1` and `mq2` are API objects for the same math field. This way a third-party script, or another module of your app, can do `var MQ = MathQuill.getInterface(3)` in its own scope, and it can do `MQ(mathFieldSpan)` to use interface version 3 to manipulate the same math field that the above example is using interface versions 1 and 2 to manipulate.
- Internally, registration of public API methods/classes (`MQ.MathField()`, `MQ.StaticMath()`) is more complicated now because `getInterface()` has to create them with a bound interface version and an isolated prototype chain (`mq1` shouldn't be `instanceof` `MQ2.MathField`)

The first breaking change using interface versioning was super easy, though, so that's a good sign: https://github.com/mathquill/mathquill/commit/2f78c8c5b5dd5ed7e4e12692759ffe0d1700e026#diff-8909a7456a5a30d96a556982356b4909

--

Though the changes are hundreds of lines, the net increase in source code is only 32 lines, which is pretty small right?
```
$ git diff --shortstat 09d8dbde723f19e80c8c362cbd0f91081ca37375^..2f78c8c5b5dd5ed7e4e12692759ffe0d1700e026^ src
 8 files changed, 277 insertions(+), 245 deletions(-)
```
(That range is from the commit where we diverge from `dev`, until 2f78c8c5b5dd5ed7e4e12692759ffe0d1700e026 which implements the first interface version check.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ for example:
 <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>
 <script src="/path/to/mathquill.js"></script>
 <script>
-  MathQuill.interfaceVersion(1);
+  MathQuill = MathQuill.getInterface(1);
   $(function() {
     MathQuill.StaticMath($('#problem')[0]);
     var answer = MathQuill.MathField($('#answer')[0], {
@@ -49,10 +49,10 @@ To load MathQuill,
 [Google CDN-hosted copy]: http://code.google.com/apis/libraries/devguide.html#jquery
 [the latest tarball]: http://mathquill.com/downloads.html
 
-To use the MathQuill API, first declare an interface version:
+To use the MathQuill API, first get the latest version of the interface:
 
 ```js
-MathQuill.interfaceVersion(1);
+MathQuill = MathQuill.getInterface(1);
 ```
 
 Now you can call `MathQuill.StaticMath()` or `MathQuill.MathField()`, which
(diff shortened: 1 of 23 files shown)


jwmerrill:
Ping @EvilOatmeal

I'm working on code review here, but it would be great to get your input on this if you have any, since you reported #464.

jwmerrill:
Done reviewing. Good to merge from my POV if you're also satisfied.

EvilOatmeal:
Looking good to me! :+1:

Issue #22, opened 2011-04-02
Typing ^ on French-Canadian-CSA keyboard


aantthony:
So to type ^ on in this layout, you press the ^-key (our "[" key) then press space.

When you press <code>x^ 2</code> it generates this:
<code>&lt;var&gt;x&lt;/var&gt;&lt;span&gt;^2&lt;/span&gt;</code>
instead of the superscript.

Using Snow Leopard 10.6.7, Safari, French-Canadian-CSA keyboard layout.

Thanks to Christian who reported this issue to me.

PR #497, opened 2016-01-19
Dead keys Safari 6

### LINKED ISSUE #205: Safari + Scandinavian keyboard = Can't type exponent
In Safari on a Scandinavian (I've tested on both Swedish and Norwegian layouts) keyboard it is not possible to use ^ to type exponents.
When I try the box appears but it's not possible to type in it, and if I move the cursor around it looks like there's something odd with it, like there's several boxes in the one box.
If I switch to an US keyboard layout it works fine. It also works fine in other browsers such as Firefox.
Here's a short video illustrating the problem: http://screencast.com/t/WRsFmcjx 
Safari version: 6.0.5 (8536.30.1)
OS X version: 10.8.5

FWennerdahl:
Fixes #205.
Might also affect #22.

This affected all dead keys such as ´,`,^,¨,~.

Tested on:
- Safari 6, 7 and 9 (OS X)
- Chrome 48 (Win and OS X)
- Firefox 42 (Win and OS X)
- IE 11
- Edge

Note that currently the Mathquill dev branch does not work in Safari 6 out of the box. Font size 0 set on the textarea (https://github.com/mathquill/mathquill/blob/dev/src/css/textarea.less#L17) prevents any input. We solved this by setting font-size to 1 and moving the hidden text-area offscreen. Don't know if this is a practical solution to implement in Mathquill, but given that the textarea uses absolute positioning it could work.

### DIFF
--- a/src/services/saneKeyboardEvents.util.js
+++ b/src/services/saneKeyboardEvents.util.js
@@ -95,6 +95,9 @@ var saneKeyboardEvents = (function() {
     var textarea = jQuery(el);
     var target = jQuery(handlers.container || textarea);
 
+    var deadkey = false;
+    var deadkeyChar = "";
+
     // checkTextareaFor() is called after keypress or paste events to
     // say "Hey, I think something was just typed" or "pasted" (resp.),
     // so that at all subsequent opportune times (next event or timeout),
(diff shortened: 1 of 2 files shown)


laughinghan:
Whoops, closed by GitHub when I deleted the dev branch, but this was good work, re-opened as #515.

PR #510, opened 2016-02-18

laughinghan:


mjdesa:
LGTM

laughinghan:
Superseded by #696, where the merge conflict is fixed.

PR #513, opened 2016-02-18

laughinghan:

[2]: https://web.archive.org/web/20150406091222/https://developers.google.com/mobile/articles/fast_buttons

(Google seems to have lost the original widely-cited "Fast Buttons" article: https://developers.google.com/mobile/articles/fast_buttons )

--

tap highlight box.


jwmerrill:
Is this ready for review? Do you want me to review it?

jwmerrill:
Reviewed the code. Looks good to me.

jwmerrill:
Rebased and opened as #622

PR #517, opened 2016-02-18
Handling of empty TextBlocks

### LINKED ISSUE #126: On dev, leaving an empty \text{} rendered from LaTeX will error

### LINKED ISSUE #313: [dev] Pasting `\text{}` causes an error

### LINKED ISSUE #429: [dev] TextBlock's latex changes to \text{undefined} after moving cursor to the left

laughinghan:
Work in progress for #474.

- Fixes #126
- Fixes #313
- Incorporates PR #430 which fixes #429


FWennerdahl:
I'll take a look at resolving the merge conflicts in this asap.

laughinghan:
You rock

laughinghan:
There's a good chance the merge conflict is just with #528, which like #430 (which is the first few commits here) fixes #429. (For why #430 wasn't merged, see here: https://github.com/mathquill/mathquill/pull/430#issuecomment-188494935 )

FWennerdahl:
I tested the master branch to verify that this PR is still relevant. When moving out of an empty text block I got

    Uncaught TypeError: Cannot read property 'nodeType' of null
which this PR adresses along with the other changes detailed in #474.

Not sure what to do about the missing CLA signing, perhaps @eoghanmcilwaine would be willing to pop by and sign it :)

In the worst case I'd have to migrate my commits to a new branch.

eoghanmcilwaine:
No worries @FWennerdahl - just signed the CLA. Sorry for the delay, I've just returned to reality after a ridiculously long air journey :sweat_smile:

FWennerdahl:
Thanks @eoghanmcilwaine! However, it seems that something happened which caused all agreements to become unsigned again (was the CLA updated? I couldn't see anything new in it).

eoghanmcilwaine:
Odd, doesn't look like it's changed. Just signed again anyway.

src/services/latex.js
```diff
@@ -2,7 +2,7 @@
 var latexMathParser = (function() {
   function commandToBlock(cmd) {
     var block = MathBlock();
-    cmd.adopt(block, 0, 0);
+    if (cmd) cmd.adopt(block, 0, 0);
```
laughinghan: minor thing: don't need this, parsers can return empty `Fragment()`s

(I'll fix)

test/unit/backspace.test.js
```diff
@@ -206,6 +206,23 @@ suite('backspace', function() {
...
+    mq.keystroke('Backspace');
+    assert.equal(cursor.parent, textBlock, 'cursor is still in text block');
+    assert.equal(cursor[R], 0, 'cursor is at the right end of the text block');
+    assert.equal(cursor[L], 0, 'cursor is at the left end of the text block');
+    assert.equal(mq.latex(), '', 'the x has been deleted');
+
+    mq.keystroke('Backspace');
+    assert.equal(cursor.parent, rootBlock, 'cursor is in root block');
+    assert.equal(cursor[L], textBlock, 'cursor is to the right of the text block');
+    assert.equal(mq.latex(), '', 'text block still empty');
```
laughinghan: wouldn't you want this backspace to delete the textblock rather than just inserting the cursor to the right? It used to be that blurring an empty textblock would delete it, that's why `deleteOutOf` just inserts the cursor right outside

laughinghan:
@eoghanmcilwaine @FWennerdahl thanks for your patience you guys! I don't know what we've done to deserve contributors like you

Was kinda cowboy about merging this with my fixes since it's actually a really easy case to stumble onto thanks to the quirky `\somethingThatsNotLatex` turning into `\text{somethingThatsNotLatex}` feature

laughinghan:
Note: merged into `master` with additional commits (b435621)

(See `git log --oneline b435621^..b435621^2`)

eoghanmcilwaine:
Nice! Thanks @laughinghan and @FWennerdahl .

PR #489, opened 2015-11-16
Support for diacritical marks + tilde

FWennerdahl:
In order to support the tilde diacritic I extracted a base class from the existing \vec implementation.

This allows for further diacritics to be added with ease as well as user-editable boxes above the text (by specifying `'&1'` as `symbol` in a DiacriticAbove binding.

I could not find any obvious location for tests of this. Seeing as \vec did not have any existing tests I omitted them for now as well. If that's a requirement I will of course write them.

The proposed CSS changes increases clarity, especially when using upper case letters such as A in the stem.

PR #518, opened 2016-02-19
Support for diacritical marks + tilde

laughinghan:
See discussion in #489. (This is the same PR but targeted at `master` instead of `dev`.)

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -457,17 +457,6 @@ LatexCmds['√'] = P(MathCommand, function(_, super_) {
   };
 });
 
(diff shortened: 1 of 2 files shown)


laughinghan:
Hmmm, the over-arrow and tilde are a little oversized, especially the arrow:

| Renderer | Screenshot |
|---|---|
| MathQuill | screen shot 2016-06-01 at 16 44 21 |
| KaTeX | screen shot 2016-06-01 at 16 44 59 |
| [MathJax](http://mathb.in/62590) | screen shot 2016-06-01 at 16 48 10 |
| LaTeXiT |  |

Eh, something to fix in the future

PR #525, opened 2016-02-23
Fix Tab while there's a selection

laughinghan:
Resolves #413

### DIFF
--- a/src/services/saneKeyboardEvents.util.js
+++ b/src/services/saneKeyboardEvents.util.js
@@ -109,7 +109,7 @@ var saneKeyboardEvents = (function() {
       clearTimeout(timeoutId);
       timeoutId = setTimeout(checker);
     }
-    target.bind('keydown keypress input keyup focusout paste', function() { checkTextarea(); });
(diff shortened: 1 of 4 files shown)


src/services/saneKeyboardEvents.util.js
```diff
@@ -109,7 +109,7 @@ var saneKeyboardEvents = (function() {
       clearTimeout(timeoutId);
       timeoutId = setTimeout(checker);
     }
-    target.bind('keydown keypress input keyup focusout paste', function() { checkTextarea(); });
+    target.bind('keydown keypress input keyup focusout paste', function(e) { checkTextarea(e); });
```
jwmerrill: NBD, but `function (e) { checkTextArea(e); }` could just be `checkTextArea`, right? Unless the are extra arguments you're trying to suppress.

jwmerrill:
Looks good to me. Feel free to make the change I suggested above or not.

src/services/saneKeyboardEvents.util.js
laughinghan: checkTextarea is set to different functions over time, the indirection is
to dereference the current `checkTextarea` function at the time the event
happens

Han

PR #526, opened 2016-02-24
Changed ** in text mode to ^

stufflebear:
^ is a more commonly used way to represent superscript


laughinghan:
and Math.js expects it and no one's using this for SymPy anyway which is why it was originally `**`

PR #528, opened 2016-02-24
Fix TextBlock cursor movement bug

### LINKED ISSUE #429: [dev] TextBlock's latex changes to \text{undefined} after moving cursor to the left

laughinghan:
Fix #429 can't move cursor left out of TextBlock.

Enforce contract that by the time `.blur()` on a node is being called, all
has been said and done and the cursor has already been moved out of the
node.

Note that the main other case that calls `.blur()` is intentional blur,
where already the cursor is hidden before `.blur()` is called on its
parent:
https://github.com/mathquill/mathquill/blob/c103063a62c85127c7eca33fa153a78a3dab6be5/src/services/focusBlur.js#L32

Supersedes and closes #516 and #430.

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -171,6 +171,10 @@ var TextBlock = P(Node, function(_, super_) {
     self.jQ[0].normalize();
 
     var textPcDom = self.jQ[0].firstChild;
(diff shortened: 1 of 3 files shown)


stufflebear:
Works for me +1

PR #536, opened 2016-03-04
Add hat command

drlippman:
Like LatexCmds.vec, this adds an accent command, \hat, for doing circumflex (hat) accents, like the Latex \hat{p}.  

Like \vec, it suffers from not being stretchy, but it works well enough for accenting single characters.


src/css/math.less
```diff
@@ -282,6 +282,18 @@
   .mq-vector-stem {
     display: block;
   }
+  
+  .mq-hat-prefix {
+    display: block;
+    text-align: center;
+    line-height: .95em;
+    margin-bottom: -.7em;
+    transform:scale(1.5,1);
```
laughinghan: the world is close but not quite ready to use only un-prefixed `transform`, especially for something like this where being backwards-compatible is harmless

http://caniuse.com/transforms2d

laughinghan: also, you know about `scaleX()`?

laughinghan:
With some vendor prefixes for the CSS transform I think this is good to go! (A matrix filter for IE would be nice someday but this fails gracefully enough there's no reason that should be a blocker)

drlippman:
I had totally forgotten about this.  Hopefully this is what you were looking for.

laughinghan:
Shoot, I'd forgotten about this too, I've been working on #646 which includes `\hat`, and also has features like adjusting height based on the size of the letter:

screen shot 2016-06-17 at 14 31 45

laughinghan:
Actually wait there's no reason for #646 to block this, this is great

PR #537, opened 2016-03-04
add autoFunctionize config option

drlippman:
This may not be something of general interest, but figured I'd share anyway.  To encourage/force my students to enter expressions that can be evaluated unambiguously, I require the use of proper function notation, like sin(x), not sinx.

To assist them with this (force it on them), this PR enables indicating certain commands that, when typed, will automatically insert parens after the command.  So typing "sin" will generate "sin()".  

This is obviously slightly annoying for things like log, where then to enter a different base you have to arrow back, but it's a tradeoff I still find valuable.

This PR adds a new config option, which works like this:

```
MQ.config({
  autoFunctionize: 'sin cos tan sec csc cot log ln'
});
```


laughinghan:
This is a pretty interesting idea, let's keep this open and see if anyone else has any interest or input?

jwmerrill:
ping @eluberoff, curious if you have any thoughts or previous experience with this idea

eluberoff:
Really like where this feature is coming from. But, in general, I find two things disconcerting when using a UI:
* when a character is inserted to the /left/ of my cursor as I'm typing (having the cursor jump just feels a little strange)
* false positive on inserting a character that then needs to be deleted

I think I'd be a little surprised by the `log(` and needing to arrow back to put in a subscript. Would be more bummed if I was trying to type cosh or sinh.

I have an alternative idea though. One interaction that has worked really well for us is to mutate the contents of an input / mathquill (to normalize it, for example) when you blur the input. The way that, for example, we harden the ghost parenthesis on blur.

I kind of love the idea of a feature where you type: `y=sin x`, blur, and it turns into `y=sin(x)`.

I'm not sure if this should live in mathquill though. Seems potentially better outside, since different systems could have different opinions about what should be auto parenthesized (y=sin2x)?

laughinghan:
I'm worried `y=sin x` &rightarrow; blur &rightarrow; `y=sin(x)` would be a lot less smooth than hardening ghost parens on blur, because ghost parens and hard parens take up the same amount of space, whereas `(x)` takes up more space than `x`, and it'd be jarring for the math to jump around on blur.

I have an even crazier, more complex, even more improbable idea: what if function operator names came with TWO ghost parens? Like, if `|` is the cursor and all these parens are gray:
```
                    lo|
         type 'g':  log(| )
        backspace:  lo|
         type 'g':  log(| )
         type 'x':  log(x|)
 ← left arrow key:  log(|x)
 ← left arrow key:  lo|g(x)
→ right arrow key:  log(|x)
         type '^':  log^{| }(x)
         type '2':  log^{2|}(x)
 ↓ down arrow key:  log^2(|x)

```
Main problem with this is that clearly with `log(x)`, if you delete `l`, `o`, or `g`, the parens should go away; but with `log(x + y)`, if you delete the `l` and the parens go away and then type `l` again where do the parens come back? Maybe if there's more than one letter, un-function-izing doesn't clear the parens. (Other special cases off the top of my head: don't insert parens if there's already parens, like with `lo|(x)` typing `g` just results in `log(|x)`.)

eluberoff:
Fair point on resizing. In Desmos, we almost always have our mathquills at
a predefined size, so the rest of the page doesn't reflow when the equation
changes. With reflowing that could be more jarring. Two ghost parens is
really interesting...

laughinghan:
I actually was talking about just within a math field. So you have `y = sin x + cos x + tan x` and you blur and the whole thing jumps

tmilev:
I am testing the autoFunctionize option and it doesn't seem to work out of the box with build 0.10.1.

Want to give you my great appreciation for your wonderful project!

I am currently writing a small computer algebra system (with LaTeX syntax). My current input is in plain text/LaTeX. For small inputs, I am about to add your wonderful editor as a parallel input interface and this option is super inportant to me because I interpret 

sin 5 x 

as 

(sin 5) x
(and I am adamant this is the correct way to interpret that notation, despite what some textbook authors appear to think).

PR #538, opened 2016-03-04
Add addCommands config option

drlippman:
I found myself wanting to add new symbols without having to modify the source code.  This PR adds a new config option for adding basic symbols (Vanilla, Binary, and Variable).

It works like this:
````
MQ.config({
 addCommands: {'oo':['VanillaSymbol','\\infty ','&infin;']}
});
````

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -767,3 +767,25 @@ var Embed = P(Symbol, function(_, super_) {
     this.latex = options.latex || noop;
   }
 });
+
+//cmds is object: {symbol:[symboltype, latex, entity]}
+//where symboltype is "VanillaSymbol", "BinaryOperator", or "Variable"
+optionProcessors.addCommands = function(cmds) {	
+  for (var str in cmds) {
+    if (LatexCmds.hasOwnProperty(str)) {
+      throw '"' + str + '" is a built-in operator name';
+    }
+    if (cmds[str].length != 3) {
+      throw '"' + str + '" does not have the required number of elements';
+    }
+    if (cmds[str][0] == 'VanillaSymbol') {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by laughinghan** [comment]
I keep wanting something like this but it feels hacky to me to hardcode VanillaSymbol, BinaryOperator, and Variable.

Also, this appears to mutate the global `LatexCmds`, so even if you did it on a single math field, it would affect all math fields. This is a problem I ran into recently with `.registerEmbed()`, which is why that's a global function :/ . Really, `LatexCmds` should probably have a system like the options objects and their prototype. Or we should just start using `Object.create()`.

File: src/commands/math/commands.js [hunk 1]
```diff
@@ -767,3 +767,25 @@ var Embed = P(Symbol, function(_, super_) {
     this.latex = options.latex || noop;
   }
 });
+
+//cmds is object: {symbol:[symboltype, latex, entity]}
+//where symboltype is "VanillaSymbol", "BinaryOperator", or "Variable"
+optionProcessors.addCommands = function(cmds) {	
+  for (var str in cmds) {
+    if (LatexCmds.hasOwnProperty(str)) {
+      throw '"' + str + '" is a built-in operator name';
```
laughinghan: Maybe this should be `has already been defined` since it may have been set by a previous `.addCommands()` call?

drlippman: Sure, I can change this if you think the approach is worth pursuing further.

drlippman:
I agree the hardcoding is hacky, but I couldn't figure any other way to do it.

I'm not sure what you mean by "Or we should just start using `Object.create()`" since I'm not that familiar with all the subsystems and approaches you're using, but I'm happy to approach `addCommands` another way if you can give me a bit more detail.

PR #539, opened 2016-03-05
Feature: .registerEmbed()

laughinghan:
<s>Usage (copied from test case):</s> (outdated, see below)
```js
    MQ.registerEmbed('thing', {
      htmlString: '<span class="embedded-html"></span>',
      text: function () { return "embedded text" },
      latex: function () { return "embedded latex" }
    });

    var mq = MQ.MathField($('<span>\\sqrt{\\embed{thing}}</span>').appendTo('#mock')[0]);
 
    assert.ok(jQuery('.embedded-html').length);
    assert.equal(mq.text(), "sqrt(embedded text)");
    assert.equal(mq.latex(), "\\sqrt{embedded latex}");
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ expose:
     latex: function() { return '\customEmbed'; }
   }
   ```
+* `ᴇxᴘᴇʀɪᴍᴇɴᴛᴀʟ` `.registerEmbed('name', function(id){return options})` allows MathQuill to parse custom embedded objects from latex, where `options` is an object like the one defined above in `.dropEmbedded`. This will parse the following latex into the embedded object you defined: `\embed{name}[id]}`
 
 [on `textarea`s]: http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-48880622
 [on `input`s]: http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-34677168
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -777,7 +777,11 @@ var Embed = LatexCmds.embed = P(Symbol, function(_, super_) {
     var self = this;
       string = Parser.string, regex = Parser.regex, succeed = Parser.succeed;
     return string('{').then(regex(/^[a-z][a-z0-9]*/i)).skip(string('}'))
-      .map(function(name) { return self.setOptions(EMBEDS[name]); })
+      .then(function(name) {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by stufflebear** [comment]
+1

**Comment by kschaefe** [comment]
This looks great.  From our standpoint, we have hacked in drop-downs inside of the MQ currently, but it isn't clean.

What is going to be the best way to replace the embed with real LaTeX?

For instance (a dumb example):
5^2\embed{dropdown with equal or not equal}25

After the user answers correctly, we really want:
5^2=25

**Comment by laughinghan** [comment]
Well in the options object you pass you may specify a function to be called
to render your embed as LaTeX, and you could return = or \ne there (for our
use case, we want it to round-trip so we're gonna return \embed{thing}, but
that's not enforced).

Han

**Comment by laughinghan** [comment]
Realized that if you have a bunch of embeds of the same type but that need to be distinguishable, this API would require you to register a whole bunch of embeds, when really you want your embed type to be parameterized on some kind of data. So, new API, `.registerEmbed()` takes in a function which will be passed `undefined` or a string:

```js
    var data;
    MQ.registerEmbed('thing', function(data_) {
      data = data_;
      return {
        htmlString: '<span class="embedded-html"></span>',
        text: function () { return "embedded text" },
        latex: function () { return "embedded latex" }
      };
    });
    var mq = MQ.MathField($('<span>\\sqrt{\\embed{thing}}</span>').appendTo('#mock')[0]);

    assert.equal(data, undefined);
 
    assert.ok(jQuery('.embedded-html').length);
    assert.equal(mq.text(), "sqrt(embedded text)");
    assert.equal(mq.latex(), "\\sqrt{embedded latex}");
 
    mq.latex('\\sqrt{\\embed{thing}[data]}');

    assert.equal(data, 'data');
 
    assert.ok(jQuery('.embedded-html').length);
    assert.equal(mq.text(), "sqrt(embedded text)");
    assert.equal(mq.latex(), "\\sqrt{embedded latex}");
```

kschaefe:
@laughinghan What's your timeframe on merging this in?  I'd be happy to test this API, as I think it will solve some of our problems, but my team's schedule will not free up for several weeks (we're having an alpha release soon).

stufflebear:
These changes still look good, but we should probably document this feature in more than this PR. I can add that

stufflebear:
Got a verbal +1 from @laughinghan

PR #542, opened 2016-03-08
[LRN] Add InertMath: non-editable with no mouse events or cursor

karoltarasiuk:
Used to render keyboard buttons so that they don't hijack
focus when they're mousdown'ed.

Conflicts:
	src/commands/math.js

### DIFF
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -466,3 +466,18 @@ API.MathField = function(APIClasses) {
     };
   });
 };
+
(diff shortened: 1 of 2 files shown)


karoltarasiuk:
This PR currently doesn't work - need to investigate - closing for now.

karoltarasiuk:
opening new PR: https://github.com/mathquill/mathquill/pull/548

PR #543, opened 2016-03-08
[LRN] Allow direct entry of $ (dollar sign)

karoltarasiuk:
By removing the binding that creates a text block
when $ is pressed.

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -287,7 +287,6 @@ var TextPiece = P(Node, function(_, super_) {
   };
 });
 
-CharCmds.$ =
 LatexCmds.text =
 LatexCmds.textnormal =
 LatexCmds.textrm =
--- a/test/unit/typing.test.js
+++ b/test/unit/typing.test.js
@@ -56,6 +56,11 @@ suite('typing with auto-replaces', function() {
       mq.typedText('\\asdf+');
       assertLatex('\\text{asdf}+');
     });
+
+    test('dollar sign', function() {
+      mq.typedText('$');
+      assertLatex('\\$');
+    });
   });
 
   suite('auto-expanding parens', function() {


laughinghan:
Thanks, I have been meaning to do this forever since almost everyone wants this off, `$` is just such a commonly used character, especially when talking about quantities.

However, I think this should probably be an option, and the option should also turn of `$` splitting apart `TextBlock`s; `\text` should either be a normal-ish command, or `$` should let you smoothly switch between text and math mode, but it shouldn't be half-way, right.

karoltarasiuk:
@laughinghan are you saying that normal $ symbol by default? and the other functionality optionally?

Also I don't exactly understand what do you mean in the second paragraph.

Now when you press $, you will switch to text mode, and everything you type until pressing $ again will be wrapped within one `\text{}` command. What would you like to change here exactly?

laughinghan:
I think either typing $ should have special behavior in both math and text, or in neither. With your change it's not special in math anymore, it's just a vanilla symbol, but if you go into text mode with `\text` and then press $, it moves you into math mode

laughinghan:
I kinda would prefer it be a config option but that might be tricky since we can't dynamically decide whether to add it to `CharCmds` or not, that happens at load-time

karoltarasiuk:
@laughinghan hey, so what do we do with this one? I'm happy to do the changes. Imo $ should work as dollar by default, and as you suggested shouldn't break out of text mode either.

You can set `CharCmds.$` to `TextBlock` on runtime as well. I pushed one more commit - you can check it. It seems to work, only when you have dollars disabled, there's no way to break from `\text`. I couldn't get `ENTER` to work as this event doesn't go that deep - any suggestions?

stufflebear:
@laughinghan says we can merge this and that the breaking from text thing can be another PR

PR #544, opened 2016-03-08
[LRN] Add extra math symbols

karoltarasiuk:
\nparallel
\measuredangle
\overrightarrow
\overleftarrow

Conflicts:
	src/css/math.less

### DIFF
--- a/src/css/math.less
+++ b/src/css/math.less
@@ -303,4 +303,28 @@
     cursor: text;
     font-family: @symbola;
   }
+
+  .mq-overarrow {
+    border-top: 1px solid black;
+    margin-top: 1px;
+    padding-top: 0.2em;
+
+    &:before {
+      display: block;
+      position: relative;
+      top: -0.34em;
+      font-size: 0.5em;
+      line-height: 0em;
+      content: '\27A4';
+      text-align: right;
+    }
+    &.mq-arrow-left:before {
+      -moz-transform: scaleX(-1);
+      -o-transform: scaleX(-1);
+      -webkit-transform: scaleX(-1);
+      transform: scaleX(-1);
+      filter: FlipH;
+      -ms-filter: "FlipH";
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


src/css/math.less
```diff
@@ -298,9 +298,27 @@
...
   }
 
+  .mq-overarrow {
+    border-top: 1px solid black;
+    margin-top: 1px;
+    padding-top: 0.2em;
 
-  &, .mq-editable-field {
-    cursor: text;
-    font-family: @symbola;
```
laughinghan: This appears to remove Symbola from all of `.mq-math-mode` (that's what the `&` refers to in this context)

laughinghan:
Thanks, this looks useful! Can you fix what appears to be a merge conflict?

karoltarasiuk:
@laughinghan yeah - I totally missed it, for some reason didn't look at the diff carefully - it's ok now

src/css/math.less
```diff
@@ -303,4 +303,28 @@
...
+      content: '\27A4';
+      text-align: right;
+    }
+    &.mq-arrow-left:before {
+      -moz-transform: scaleX(-1);
+      -o-transform: scaleX(-1);
+      -webkit-transform: scaleX(-1);
+      transform: scaleX(-1);
+      filter: FlipH;
+      -ms-filter: "FlipH";
```
laughinghan: LOL

stufflebear:
+1

src/css/math.less
karoltarasiuk: Believe me or not, we couldn't find the opposite arrow which looked exactly the same...

This is the right one: http://www.fileformat.info/info/unicode/char/27a4/index.htm It's called BLACK RIGHTWARDS ARROWHEAD. LEFTWARDS one doesn't exist...

laughinghan: I heard they considered including it, but were afraid there wasn't enough
space for the taco emoji

Han

karoltarasiuk: LOL that would be it I bet!

PR #545, opened 2016-03-08
[LRN] Automatically add sub block for limits

karoltarasiuk:
Conflicts:
    src/commands/math/basicSymbols.js
    test/unit/typing.test.js

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -161,7 +161,7 @@ var AutoOpNames = Options.p.autoOperatorNames = { _maxLength: 9 }; // the set
   // of operator names that MathQuill auto-unitalicizes by default; overridable
 var TwoWordOpNames = { limsup: 1, liminf: 1, projlim: 1, injlim: 1 };
 (function() {
-  var mostOps = ('arg deg det dim exp gcd hom inf ker lg lim ln log max min sup'
+  var mostOps = ('arg deg det dim exp gcd hom inf ker lg ln log max min sup'
                  + ' limsup liminf injlim projlim Pr').split(' ');
   for (var i = 0; i < mostOps.length; i += 1) {
     BuiltInOpNames[mostOps[i]] = AutoOpNames[mostOps[i]] = 1;
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -311,7 +311,31 @@ LatexCmds['^'] = P(SupSub, function(_, super_) {
   };
 });
 
-var SummationLimitNotation = P(MathCommand, function(_, super_) {
+var SummationNotation = P(MathCommand, function(_, super_) {
+  _.init = function(ch, html) {
+    var htmlTemplate =
+      '<span class="mq-large-operator mq-non-leaf">'
+    +   '<span class="mq-to"><span>&1</span></span>'
+    +   '<big>'+html+'</big>'
+    +   '<span class="mq-from"><span>&0</span></span>'
+    + '</span>'
+    ;
+    Symbol.prototype.init.call(this, ch, htmlTemplate);
+  };
+  _.latex = function() {
+    function simplify(latex) {
+      return latex.length === 1 ? latex : '{' + (latex || ' ') + '}';
+    }
+    return this.ctrlSeq + '_' + simplify(this.ends[L].latex()) +
+      '^' + simplify(this.ends[R].latex());
+  };
+  _.createLeftOf = function(cursor) {
+    super_.createLeftOf.apply(this, arguments);
+    if (cursor.options.sumStartsWithNEquals) {
+      Letter('n').createLeftOf(cursor);
+      Equality().createLeftOf(cursor);
+    }
+  };
   _.parser = function() {
     var string = Parser.string;
     var optWhitespace = Parser.optWhitespace;
@@ -337,33 +361,6 @@ var SummationLimitNotation = P(MathCommand, function(_, super_) {
     this.upInto = this.ends[R];
     this.ends[L].upOutOf = this.ends[R];
     this.ends[R].downOutOf = this.ends[L];
-  };
-});
-
-var SummationNotation = P(SummationLimitNotation, function(_, super_) {
-  _.init = function(ch, html) {
-    var htmlTemplate =
-      '<span class="mq-large-operator mq-non-leaf">'
-    +   '<span class="mq-to"><span>&1</span></span>'
-    +   '<big>'+html+'</big>'
-    +   '<span class="mq-from"><span>&0</span></span>'
-    + '</span>'
-    ;
-    Symbol.prototype.init.call(this, ch, htmlTemplate);
-  };
-  _.latex = function() {
-    function simplify(latex) {
-      return latex.length === 1 ? latex : '{' + (latex || ' ') + '}';
-    }
-    return this.ctrlSeq + '_' + simplify(this.ends[L].latex()) +
-      '^' + simplify(this.ends[R].latex());
-  };
-  _.createLeftOf = function(cursor) {
-    super_.createLeftOf.apply(this, arguments);
-    if (cursor.options.sumStartsWithNEquals) {
-      Letter('n').createLeftOf(cursor);
-      Equality().createLeftOf(cursor);
-    }
   };
 });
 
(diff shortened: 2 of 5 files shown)


src/commands/math/basicSymbols.js
```diff
@@ -140,7 +140,7 @@ var BuiltInOpNames = {}; // http://latex.wikia.com/wiki/List_of_LaTeX_symbols#Na
 var TwoWordOpNames = { limsup: 1, liminf: 1, projlim: 1, injlim: 1 };
 var AutoOpNames = Options.p.autoOperatorNames = { _maxLength: 9 };
 (function() {
-  var mostOps = ('arg deg det dim exp gcd hom inf ker lg lim ln log max min sup'
+  var mostOps = ('arg deg det dim exp gcd hom inf ker lg ln log max min sup'
```
laughinghan: Oh, this is an unfortunate tradeoff I hadn't considered. So this now requires `lim` to be an `autoCommand` or you have to type a backslash? If it's an `autoCommand`, it has the unfortunate result that typing `l`, `i`, `m`, and `Backspace` deletes everything rather than leaving you with `li`, right? I guess `sum` as an `autoCommand` has the same problem, but at least ∑ doesn't look like 3 separate characters.

I think I was originally imagining that `lim` would still auto-unitalicize normally, and to get the special approaches blocks you'd type `_` after it, how would you feel about that? (When the special approaches blocks are active, the cursor can't go between the letters; but once you delete it, it'd go back to being 3 letters that happen to be unitalicized, just like any other auto-unitalicized command.)

src/css/math.less
```diff
@@ -298,6 +298,21 @@
     }
   }
 
+  .mq-limit {
+    text-align: center;
+
+    span  {
+      display: block;
```
laughinghan: this is causing every `span` in the limit's "approaches" block to be on its own line: 
image

karoltarasiuk: yeah, that's correct, and this is something I noticed with a cursor as well... will fix it

src/commands/math/basicSymbols.js
karoltarasiuk: In our approach you need to type a backslash and then lim: `\lim`, after you press enter it moves you automatically below the lim, and then you can type. Pressing right arrow takes you out of the command.

src/css/math.less
karoltarasiuk: @laughinghan below added `span > span` rule to fix this issue. We had it in our codebase but on a separate commit, which I missed

src/commands/math/basicSymbols.js
karoltarasiuk: @laughinghan any suggestion here or my explanation sounds good?

laughinghan: so in your approach what does just plain typing `l`, `i`, and `m` do? They remain italicized, like 3 variables?

src/css/math.less
laughinghan: oh, this is so that the `.mq-approaches` will be below the `lim`? How about instead of setting `display: block` on every span and then overriding that, just add `.mq-lim` class and select for that?

src/commands/math/basicSymbols.js
karoltarasiuk: Yep, exactly:

src/css/math.less
karoltarasiuk: yeah, it's better, the class was already on this element, so I just changed CSS, it looks like this now: 
screen shot 2016-03-10 at 5 12 57 pm

karoltarasiuk: now I'm taking it back, as there can be other symbols re-using those classes... I should come up with sth unique

src/commands/math/basicSymbols.js
laughinghan: yep exactly haha `\lim_{\sin}`

src/css/math.less
laughinghan: yep exactly, like `\lim_{\sin}` haha

karoltarasiuk: ok I pushed it, but I don't really get your last example - it seemed to work fine

laughinghan: my mistake, `.mq-operator-name` used to be `.mq-un-italicized`

src/commands/math/commands.js
```diff
@@ -314,31 +314,7 @@ LatexCmds['^'] = P(SupSub, function(_, super_) {
...
-      Equality().createLeftOf(cursor);
-    }
-  };
-  _.latex = function() {
-    function simplify(latex) {
-      return latex.length === 1 ? latex : '{' + (latex || ' ') + '}';
-    }
-    return this.ctrlSeq + '_' + simplify(this.ends[L].latex()) +
-      '^' + simplify(this.ends[R].latex());
-  };
```
laughinghan: there's a bug in this parsing, it produces 2 child blocks for both sums and limits, you can see this bug by doing:
```js
mathField.latex('\\lim').keystroke('Left').typedText('x').latex()
```
You should get `\lim_x`, but you get `\lim_{ }` instead. You'll find that any `\lim` rendered from LaTeX, moving around with the left and right arrow keys is messed up because of the invisible second block (what in summation notation is the upper limit block).

Did you guys do this factoring this way because you it was too difficult to figure out how to write your own parser? We've gotten a lot of feedback that it's quite inscrutable, is that your experience as well?

laughinghan: (I'll fix it, btw)

laughinghan: @karoltarasiuk Suggested fix: https://github.com/Learnosity/mathquill/pull/48

src/commands/math/basicSymbols.js
laughinghan: I'd prefer not to remove `lim` from the auto-operator names list, since I think the original reason it was there still holds (namely that if you saw `lim a_n` in math, even if `lim` is italicized, you would read it as "limit of `a_n`"). Is the concern here that it won't round-trip? That typing `l`, `i`, and `m` will export as in `\lim`, which when imported becomes `\lim_{ }`? Potentially I could update https://github.com/Learnosity/mathquill/pull/48 so that `\lim` without a subscript afterwards is parsed into the auto-operator name (3 separate letters) instead of into `\lim_{ }`, and only `\lim_x` (subscript in the original LaTeX) has the underset subscript, and we could see if we like that better.

karoltarasiuk:
@laughinghan I rebased this branch on top of master and cherry-picked your commit with proposed changes.

Now about removing it from operator list: if we don't do that this snippet of code will simply do nothing:
```
// Add mathquill auto-commands (space-delimited list of latex commands)
mathquill.config({ autoCommands: 'lim'});
```
What we do here is enabling auto converting `lim` to `\lim` and moving you to the subscript position. Is there any other way to achieve that? If yes why not - we can add it back to operator list.

laughinghan:
This looks great, the auto-converting `lim` to `\lim` and moving to subscript position works well. Remaining fixes:
- if you have `li|` and type a single `m` and get `\lim_{| }`, then a single backspace should become `li|` again rather than deleting the whole thing
- backspacing or fwd-deleting `\lim_{stuff}` should move the cursor into the appropriate end of the subscript block, just like backspacing or fwd-deleting into `^{stuff}`

Also, an idea: what if instead of typing `m` triggering the underneath subscript, what if `lim` was merely upright, and typing an underscore `_` after that is what triggers the underneath subscript? Do people ever do limits without the thing?

karoltarasiuk:
@laughinghan `lim` is always with an underneath subscript, hence it converts right after you type `m`, because there is no other valid use case. I will try to find some time to work on suggested fixes.

CoryGH:
It's been over a year, is this going to get merged?

CoryGH:
I'd like to recommend this `css` change since it looks a bit nicer:

```
.mq-limit {
  text-align: center;
  vertical-align: middle;
}
.mq-limit .mq-lim {
  display: block;
  font-size: 70%;
}
.mq-limit .mq-approaches {
  font-size: 50%;
  float: right; /* take out of normal flow to manipulate baseline */
  width: 100%;
  display: block;
}
```

qadeesz:
> .mq-limit {
>   text-align: center;
>   vertical-align: middle;
>   margin: 2px;
> }
> .mq-limit .mq-lim {
>   display: block;
>   font-size: 70%;
> }
> .mq-limit .mq-approaches {
>   font-size: 50%;
>   float: right; /* take out of normal flow to manipulate baseline */
>   width: 100%;
>   display: block;
> }
> ```
> 
> For example:
> 
> `\int _a^{\infty }f\left(x\right)dx=\lim _{b\to \infty }\int _a^bf\left(x\right)dx`
> 
> Produces:
> 
> image

not work for me how work this?

PR #547, opened 2016-03-08
Add getSelection API function

drlippman:
I was trying to figure out how to use the API to wrap the current selection with an expression larger than one symbol, like wrapping the selection in `\\sin\right( [selection] \left) `, and couldn't figure a way.  

This PR adds a new API function to return the current selection will allow users to utilize the current selection, then update it with an API .write() command.

### DIFF
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -179,6 +179,13 @@ function getInterface(v) {
       this.__controller.cursor.clearSelection();
       return this;
     };
+    _.getSelection = function() {
+      if (this.__controller.cursor.selection) {
+        return this.__controller.cursor.selection.join('latex');
+      } else {
+	return null;
+      }
+    };
 
     _.moveToDirEnd = function(dir) {
       this.__controller.notify('move').cursor.insAtDirEnd(dir, this.__controller.root);


laughinghan:
Interesting. When you do that, you wind up with the cursor to the right of the close-paren, right? Is that what you want?

I've been thinking of extending our input LaTeX syntax with a `\MathQuillCursor` command which would mark where to put the cursor or the current selection, so you could do like `mathField.write('\\sin\\left( \MathQuillCursor \\right)')` and if there was a current selection, it would get wrapped and the cursor would be at the end _inside_ the close paren. Does that sound like it would work better for you?

Are there any other use cases for `.getSelection()`?

drlippman:
Yes, the cursor ends up to the right of the close-paren, but a call to `.keystroke('Left')` fixes that.

Having a `\MathQuillCursor` command to place the cursor or current selection would be lovely, and would address my use case.  

I can't think of any other particular use cases for `.getSelection()`.

thatcomputerguy0101:
I am trying to utilize text selection as a way of hinting to which part of an equation a user intends to interact with. (In a context where equations are not directly edited.) To do so, I patched on a read-only selection API (that could be possibly extended to also have write capabilities) that exposes a Selection object with six properties: the original LaTeX expression, a start and end index for the selection, the slice of the LaTeX that is selected, the whole LaTeX expression with `\MathQuillSelectionStart` and `\MathQuillSelectionEnd` inserted at the selection bound (always valid LaTeX, and the whole expression with the selection portion wrapped with `\MathQuillSelection{...}` (invalid if it crosses another grouping boundary). If there is still interest in this functionality, I can put together a PR for it.

My existing implementation is at [thatcomputerguy0101/PhysicsNotebook .../selection.js](https://github.com/thatcomputerguy0101/PhysicsNotebook/blob/main/js/manipulations/selection.js). It also has a seventh output using `\operatorname{MathQuillSelection}` for compatibility with stricter LaTeX parsers, but that can be easily derived from one of the other outputs so I don't feel that it is necessary to keep. I would also update the names used for each of the properties, as the current ones were developed organically instead of being thought out.

PR #548, opened 2016-03-08
[LRN] Add InertMath: non-editable with no mouse events or cursor

karoltarasiuk:
Used to render keyboard buttons so that they don't hijack
focus when they're mousdown'ed.

Conflicts:
    src/commands/math.js

### DIFF
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -466,3 +466,22 @@ API.MathField = function(APIClasses) {
     };
   });
 };
+
+/**
+ * This is almost the same as StaticMath, except we don't
+ * bind any mouse events (which give us selectability and move
+ * around an invisible cursor).
+ *
+ * Otherwise, if you have a button that contains a rendered StaticMath symbol
+ * which causes something to be written into an editable mathquill,
+ * the mouseup event will hijack focus.
+ */
+API.InertMath = function(APIClasses) {
+  return P(APIClasses.AbstractMathQuill, function(_, super_) {
+    this.RootBlock = MathBlock;
+    _.__mathquillify = function() {
+      super_.__mathquillify.call(this, 'mq-math-mode');
+      return this;
+    };
+  });
+};
--- a/src/commands/math.js
+++ b/src/commands/math.js
@@ -483,5 +483,18 @@ API.InertMath = function(APIClasses) {
       super_.__mathquillify.call(this, 'mq-math-mode');
       return this;
     };
+    _.init = function() {
+      super_.init.apply(this, arguments);
+      this.__controller.root.postOrder(
+        'registerInnerField', this.innerFields = [], APIClasses.MathField);
+    };
+    _.latex = function() {
+      var returned = super_.latex.apply(this, arguments);
+      if (arguments.length > 0) {
+        this.__controller.root.postOrder(
+          'registerInnerField', this.innerFields = [], APIClasses.MathField);
+      }
+      return returned;
+    };
   });
 };
--- a/test/demo.html
+++ b/test/demo.html
@@ -61,6 +61,7 @@ code span {
 
 <p>
   Insert using inertMath <button class="insert-trigger" data-latex="\sqrt{4}"><span class="mathquill-inert-math">\sqrt{4}</span></button> inside formulas below:<br>
+  Insert using inertMath with innerFields <button class="insert-trigger" data-latex="\sqrt{4}\frac{1}{\MathQuillMathField{\sqrt{4}}}"><span class="mathquill-inert-math">\sqrt{4}\frac{1}{\MathQuillMathField{\sqrt{4}}}</span></button> inside formulas below:<br>
   <span class="mathquill-math-field"></span><br>
   or here: <span class="mathquill-math-field">\frac{2}{3}</span><br>
   Now insert using staticMath <button class="insert-trigger" data-latex="\sqrt{4}"><span class="mathquill-static-math">\sqrt{4}</span></button> to the same formulas above, and notice that you can select all the text within a button.<br>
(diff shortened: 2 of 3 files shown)


src/commands/math.js
```diff
@@ -466,3 +466,23 @@ API.MathField = function(APIClasses) {
     };
   });
 };
+
+/**
+ * This is almost the same as StaticMath, except we don't
+ * bind any mouse events (which give us selectability and move
+ * around an invisible cursor).
```
laughinghan: ah, this makes sense

src/commands/math.js
```diff
@@ -466,3 +466,23 @@ API.MathField = function(APIClasses) {
...
+ * Otherwise, if you have a button that contains a rendered StaticMath symbol
+ * which causes something to be written into an editable mathquill,
+ * the mouseup event will hijack focus.
+ */
+API.InertMath = function(APIClasses) {
+  return P(APIClasses.StaticMath, function(_, super_) {
+    this.RootBlock = MathBlock;
+    _.__mathquillify = function() {
+      super_.__mathquillify.call(this, 'mq-math-mode');
+      this.__controller.staticMathTextareaEvents();
```
laughinghan: as far as I know, `staticMathTextareaEvents()` is part of making mouse selection of StaticMath's work (that is, after mouse selecting, it lets you Cmd-C to copy the selection), and doesn't do anything else, so I don't think you even need this

test/demo.html
```diff
@@ -59,6 +59,18 @@
...
 
+<p>
+  Insert <button class="insert-trigger" data-latex="\sqrt{4}"><span class="mathquill-inert-math">\sqrt{4}</span></button> inside formulas below:<br>
+  <span class="mathquill-math-field"></span><br>
+  or here: <span class="mathquill-math-field">\frac{2}{3}</span>
+</p>
+
+<p>
+  But if you won't use InertMath, the result will be far from perfect:<br>
+  Insert <button class="insert-trigger" data-latex="\sqrt{4}"><span class="mathquill-math-field">\sqrt{4}</span></button> inside formulas below:<br>
```
laughinghan: Why do you have an editable math field inside a button?

test/demo.html
```diff
@@ -116,6 +129,14 @@
     latexMath.latex(decodeURIComponent(location.hash.slice(1))).focus();
 });
 
+$(".insert-trigger").click(function () {
+  var latex = $(this).data('latex');
+  $(this).parent().find('.mathquill-math-field').each(function () {
+    var mathquill = MQ(this);
+    mathquill.write(latex);
+  });
```
laughinghan: this appears to seek out every editable math field in the paragraph and write `\sqrt{4}` into it, including the editable math field in the button, is this not expected behavior?

test/demo.html
```diff
@@ -59,6 +59,18 @@
 
 <p>This button runs the JavaScript code written on it to MathQuill-ify the following <code>&lt;span&gt;</code> element into an editable math textbox: <button onclick="$(this).text('MQ(this.nextSibling).revert()'); MQ.MathField(this.nextSibling); var orig = arguments.callee; this.onclick = function(){ $(this).text('MQ.MathField(this.nextSibling)'); MQ(this.nextSibling).revert(); this.onclick = orig; };">MQ.MathField(this.nextSibling)</button><span>\frac{d}{dx}\sqrt{x} = \frac{d}{dx}x^{\frac{1}{2}} = \frac{1}{2}x^{-\frac{1}{2}} = \frac{1}{2\sqrt{x}}</span></p>
 
+<p>
+  Insert <button class="insert-trigger" data-latex="\sqrt{4}"><span class="mathquill-inert-math">\sqrt{4}</span></button> inside formulas below:<br>
```
laughinghan: Using a StaticMath (`mathquill-static-math`) seems to work equally as well as an InertMath in this case?

laughinghan:
I think this could be a useful config option on static math instances, but I don't think this ought be its own top-level mathquill API class, and static math also doesn't seem to have the problem you're fixing (perhaps anymore)?

src/commands/math.js
karoltarasiuk: makes sense updating

karoltarasiuk: Actually I could remove this, but it wasn't fixing selectability issues. And I noticed that I shouldn't inferit from StaticMath but from AbstractMathQuill, and then it works.

test/demo.html
karoltarasiuk: Yep, see above - I made a mistake and correct it now.

test/demo.html
karoltarasiuk: Removed this - was rubbish indeed. I edited previous example to make it more meaningful.

test/demo.html
karoltarasiuk: ^^^

karoltarasiuk:
@laughinghan added changes, as really it wasn't working as expected because of a bug in the code. Now it works as expected and you can really see it in the demo page.

Static math is selectable, inert is not - I'd still argue that both are useful. Could introduce a flag like `ignoreInteractions` and then load StaticMath with a lot of `if`s in there. Don't think it would be ideal as now code looks very neat and tidy. But your call after all.

laughinghan:
Actually I like this so much it should be the default behavior for StaticMath in the first place! :) It's basically what KaTeX does.

I think it's actually just one `if` statement, the other thing that StaticMath does that InertMath doesn't is `.innerFields`, which I don't think you're deliberately disabling, right?

Since it's a breaking change I'll submit a little PR to add it using the API versioning scheme

kschaefe:
We use StaticMath with `.innerFields`.  Just confirming that you're not removing that functionality.

laughinghan:
I'm not! I'm not removing `.innerFields`, I think that's an important feature that InertMath needs too

kschaefe:
Great! It was late last night and I think my paranoia must have been getting the best of me.

karoltarasiuk:
@laughinghan I've added `.innerFields` on a separate commit, but I don't think they're valid use case here... I've added changes to the demo example as well, and you can see, that having `.innerFields` in InertMath can be rather pointless. Unless you don't want them to be editable, but only renderable.

kschaefe:
We just came across a bug in our software that would be solved by this.  What's the status on getting this merged?

laughinghan:
Well I don't want a separate `InertMath`, I want this to be a config option on `StaticMath`.

Also,

> having `.innerFields` in InertMath can be rather pointless. Unless you don't want them to be editable, but only renderable.

this seems surprising to me, this is def not what I'd want, what I'd expect is for the inner fields to be editable, but selecting the inert math does normal browser selection rather than MathQuill selection.

Maybe they appear inert because they don't respond to clicks, because inner editable fields in an inert static math need to notice it's inert and call `.delegateMathFields()` on themselves (the inner editable fields)?

Issue #11, opened 2010-12-20
Shortcuts


magicvinni:
Hello, 
At first congratulations for this wonderfull code. I was looking for a such component since a long time. I'm a french math teacher (So excuse my bad english) and I'll use it in my web page for my students.

It 'd be great to have shorcuts (ex : CTRL+R or ALT+R for \sqrt ....). And if the list of specials keys could be in an external (editable) file, it will be better

At the time, is it possible to have a function like insert(id,what) which insert the text "what" at the cursor place in span "id".

Thanks a lot !
Vincent

PR #549, opened 2016-03-08
Add keyboard passthrough for substituteTextarea

drlippman:
When using substituteTextarea with a <span tabindex="0"> element to prevent a mobile onscreen keyboard from displaying, the downside is that the regular keyboard doesn't work either.

This PR adds another config option: `keyboardPassthrough: true`, which listens for keypresses on the span element and passes them through to the MathField's keystroke handler. This seems to work fine for basic characters and control sequences.

For Ctrl-C, Ctrl-X, and Ctrl-V copy/paste, to avoid compatibility issues an internal clipboard is used, allowing copy/pasting within and between MathQuill areas, but not from/to other textfields. 

I don't know if this is the best approach, but it seems to work OK.

### DIFF
--- a/src/services/saneKeyboardEvents.util.js
+++ b/src/services/saneKeyboardEvents.util.js
@@ -91,7 +91,8 @@ var saneKeyboardEvents = (function() {
   return function saneKeyboardEvents(el, handlers) {
     var keydown = null;
     var keypress = null;
-
+    var usedkeydown = null;
+    
     var textarea = jQuery(el);
     var target = jQuery(handlers.container || textarea);
 
@@ -147,7 +148,7 @@ var saneKeyboardEvents = (function() {
     function onKeydown(e) {
       keydown = e;
       keypress = null;
-
+      usedkeydown = false;
       if (shouldBeSelected) checkTextareaFor(function(e) {
         if (!(e && e.type === 'focusout') && textarea[0].select) {
           textarea[0].select(); // re-select textarea in case it's an unrecognized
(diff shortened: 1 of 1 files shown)


laughinghan:
Oh interesting, Desmos actually substitutes a `<textarea readonly>` to suppress the onscreen keyboard, but then if a `keydown` event happens, sets `.readOnly = false`. (Except for iOS, where a `<textarea readonly>` has the undesirable property of bringing up an annoying focus bar thing at the bottom yet lacks the required property of firing `keydown` events if you type on a hardware keyboard attached via Bluetooth. And they treat Android the same as iOS although I don't think we did extensive testing on that front.)

MathQuill actually started out [with `tabindex=0` on its containing span](https://github.com/mathquill/mathquill/commit/4c417f89fcdb1546ba92129c6a2474ba5195ee62#diff-e2a2dd38d8f45ec22cec9654fb488424L716) instead of a hidden textarea and [did `String.fromCharCode(e.which)` on `keypress`](https://github.com/mathquill/mathquill/commit/aceadb006186eea4a9d9555ac73d49e6f182cabe#diff-6fcb8bd3232471c30d0ec6b0299630a0L348), so it's funny to see this come full circle.

From what I recall, one of the reasons we switched to a real textarea was stuff like, backslash on a French keyboard didn't work (backslash is AltGraph+8 on a French keyboard, apparently), which you'd think is a super edge case, but I guess not for our users because it's the second Issue ever opened on MathQuill by someone other than myself, #11!

There are edge cases on US keyboards, too, that don't work with the `fromCharCode` technique. For example, you're on a Mac, right? Alt+`<` will type `≤` on a Mac, but is broken by `keyboardPassthrough` (just tried it). With Desmos' blessing I'll share the snippet that does the `readOnly` swapping, IIRC it has fewer problems.

jwmerrill:
Yes, feel free to share/document that if you like.

drlippman:
I guess I shouldn't be surprised you already tried the same thing :)  I'd love to see the `<textarea readonly>` approach, but the lack of support for attached keyboards is unfortunate.  Since my usecase is a bit more limited, the tradeoff of not handling some special keystrokes might be worth having attached keyboards work (which I haven't actually tested yet)

drlippman:
I made a few tweaks to prevent blocking of modifier+key combos, like `Alt <` and `Ctrl-Alt-8`, which produce keypress events that seem to work fine when passed to `.typedText()` .  Now the Mac `≤` seems to work, along with the backslash and other combinations on a French keyboard layout. 

If there are other edge cases to worry about, or other reasons I should abandon this approach, please let me know.

PR #550, opened 2016-03-08
Fix optional data argument to \embed

laughinghan:
Somehow test was broken and neither of us noticed?

Be a little more forgiving for data than for embed name, allow letters
digits dashes underscores and whitespace


stufflebear:
Looks good to me. I can't wait until we have CI

PR #551, opened 2016-03-08
Upgrade to Mocha 2.4 to fix iffy "stealing back" focus test

laughinghan:
Recommend -w to view diff.

That test (from #355) requires the document to be focused for the browser to exhibit
the "stealing back" focus behavior, so we had a check to skip the test
if the browser isn't focused.

Unfortunately that check was useless a large fraction of the time due to
a race condition: if you Refresh unit.html, while tests are all being
defined but none are being run the check happens, and then while the
whole test suite is actually running (slow), you switch to another
browser tab, then when the test actually runs the document is unfocused
and the test fails.

(Note that this only happens if you have the Developer Tools/Web
 Inspector closed; when they're open and focused the document will be
 unfocused anyway, even when you Refresh. So to reproduce that above
 behavior make sure Dev Tools aren't open on unit.html when you Refresh.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "pjs": "3.x"
   },
   "devDependencies": {
-    "mocha": "*",
+    "mocha": ">=2.4.1",
     "uglify-js": "2.x",
     "less": ">=1.5.1"
   }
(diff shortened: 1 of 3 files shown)


laughinghan:
Oh oops, and the reason Mocha v2.4 is required is that `this.skip()` broke the HTML test reporter until v2.4: https://github.com/mochajs/mocha/issues/1760#issuecomment-170400139

PR #552, opened 2016-03-09
[LRN] Added to new symbols

karoltarasiuk:
- circled dot
- parallelogram

Conflicts:
	test/unit/latex.test.js


stufflebear:
Works for me
screenshot 2016-03-08 16 54 20

karoltarasiuk:
Yeah - it's `\odot` in MathJax. We got following macros there:
```
circledot: ["{\\odot}"],
parallelogram: ["\\unicode{x25B1}"],
```

We can keep `\circledot` compatibility internally on our fork. We've added it on client's request, so we'll have to keep it for backwards compatibility.

`\parallelogram` is a different thing, and imo worth having, as there is no other command. Also the name will unlikely be needed for any other symbol. I put a comment as you suggested.

laughinghan:
`\circledot` doesn't conflict with any real commands (right?) so feel free to do `LatexCmds.odot = LatexCmds.circledot = bind(VanillaSymbol, ...`, it seems perfectly reasonably something someone would try typing.

karoltarasiuk:
Yeah - that's what I was planning to do internally, as when adding it I missed the fact that MathQuill already had `\odot`. OK then - then changes done and PR updated.

laughinghan:
:sparkles:

PR #553, opened 2016-03-09
[LRN] Add `empty()` public method and `clear()` as its alias

karoltarasiuk:
Removes all math from the editable field

Conflicts:
    src/commands/math.js
    src/commands/math/commands.js
    src/publicapi.js

### DIFF
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -158,6 +158,10 @@ function getInterface(v) {
       if (this.__controller.blurred) this.__controller.cursor.hide().parent.blur();
       return this;
     };
+    _.empty = _.clear = function() {
+      this.select().__controller.backspace();
+      return this;
+    };
     _.cmd = function(cmd) {
       var ctrlr = this.__controller.notify(), cursor = ctrlr.cursor;
       if (/^\\[a-z]+$/i.test(cmd)) {
--- a/test/unit/publicapi.test.js
+++ b/test/unit/publicapi.test.js
@@ -114,7 +114,7 @@ suite('Public API', function() {
       mq.latex('x+y');
       assert.equal(mq.html(), '<var>x</var><span class="mq-binary-operator">+</span><var>y</var>');
     });
-    
+
     test('.text() with incomplete commands', function() {
       assert.equal(mq.text(), '');
       mq.typedText('\\');
@@ -124,7 +124,7 @@ suite('Public API', function() {
       mq.typedText('qrt');
       assert.equal(mq.text(), '\\sqrt');
     });
-    
+
     test('.text() with complete commands', function() {
       mq.latex('\\sqrt{}');
       assert.equal(mq.text(), 'sqrt()');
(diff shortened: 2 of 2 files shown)


src/publicapi.js
```diff
@@ -158,6 +158,10 @@ function getInterface(v) {
       if (this.__controller.blurred) this.__controller.cursor.hide().parent.blur();
       return this;
     };
+    _.empty = _.clear = function() {
+      this.select().__controller.backspace();
+      return this;
+    };
```
laughinghan: Can we not have `.clear()` as an alias? API methods are a bit more of a liability than LaTeX commands.

Also, I'd expect this to be an inefficient way to empty the root (`.select()` in particular is relatively expensive since it wraps things in a `<span class="selection">` and writes the LaTeX into a textarea and selects it (since that's how cut and copy work)), I'd recommend something like what `renderLatexMath` does if passed an empty LaTeX, which is basically 5 lines, https://github.com/mathquill/mathquill/blob/master/src/services/latex.js#L108-L109 and https://github.com/mathquill/mathquill/blob/master/src/services/latex.js#L124-L128

eoghanmcilwaine: Yeah, the `.select()` thing isn't ideal - we must come up with a better approach.

@karoltarasiuk was there a reason for the double `.empty()` / `.clear()` naming? We only need one method name on our side.

Just to add some context @laughinghan: we added this method so we could clear just the currently focused `\MathQuillMathField{}` without affecting stuff around it. If there were no `\MathQuillMathField{}`s, we wouldn't need this method as we could just call `.latex('')`.

laughinghan: Well if you do \MathQuillMathField[answer]{}, could you can do
outerMathFieldVariableNameWhatever.innerFields.answer.latex('')?

Han

eoghanmcilwaine: Is there a way to access whichever \MathQuillMathField currently contains the cursor?

laughinghan: I don't think so, I could see that being useful

laughinghan: `MQ($(document.activeElement).closest('.mq-editable-field')[0])` works but is kinda brittle

eoghanmcilwaine: Could be less of a hack than what we did :grinning: 

I think that [accessing the focused field] was one of the things we tried to add in the [StaticMathWithEditables commit](https://github.com/Learnosity/mathquill/commit/5c80199f028f7edd5a1ed5fe389764e0b1601ef0) that @karoltarasiuk  mentioned on our call the other night (that's what `withActiveNode()` is there for).

What we did isn't a great solution - because sometimes you want your method to act on the entire mathquill instance, and other times you want it to affect just the focused field. E.g. I can't just change the `.latex()` method so it only affects the focused bit.

eoghanmcilwaine: Actually another issue was: if you've got multiple \MathQuillMathFields but none of them are currently focused, which one do you programmatically write into? (Or affect with stuff like `.clear()`). 

So now we're remembering the last focused input (by adding a `.mq-last-focused` class)

laughinghan:
Merged with modifications as #750

Issue #310, opened 2014-05-14
[dev] Both native and MathQuill cursors are visible on iOS


alopatin:
As of the latest commit (1e6aa06e6c) both the native and MathQuill cursors are visible when you focus into a MathQuill input on iOS, which is visually confusing:

![mq](https://cloud.githubusercontent.com/assets/238458/2978355/830c0290-dbb8-11e3-8f72-67a0a98c964b.gif)

The above gif is Chrome on iOS 7, but Safari shows the same behavior as well. Haven't tested any Android browsers.

This is a blocker for deploying a MathQuill integration to mobile web users.

jsg2021:
its visible in iOS9.

PR #554, opened 2016-03-09
[LRN] Add defensive CSS for hidden textarea

karoltarasiuk:


src/css/textarea.less
```diff
@@ -6,6 +6,11 @@
 
     // TODO: why is this here?
     .user-select(text);
+
+   > textarea {
+     opacity: 0; // Defense against iOS cursor appearing in client pages
```
laughinghan: Is this to fix #310? Should be fixed already (https://github.com/mathquill/mathquill/pull/554/files#diff-f00a1908060a90535e10dbd689e744ddR22)

laughinghan:
Also, we prefer to have these in `.mq-textarea *` rather than `.mq-textarea > textarea` because we let people substitute something else with the `substituteTextarea` config option. (People substitute stuff like spans with a tabindex attribute (which makes it focusable) in order to disable the virtual touchscreen keyboard.)

src/css/textarea.less
karoltarasiuk: hey @keremdaldal, do you remember whether this is the very issue?

keremdaldal: @karoltarasiuk, @laughinghan, yes it's the same issue. You can safely remove this line.

src/css/textarea.less
```diff
@@ -6,6 +6,11 @@
...
+     overflow: hidden; // Defense against bug on Chrome with retina
```
karoltarasiuk: @keremdaldal do you remember what was this one?

src/css/textarea.less
karoltarasiuk: good stuff thanks! I will update PR, once we solve the other one

karoltarasiuk:
Spoken to @keremdaldal and we decided to close it as it seems those fixes are not needed anymore.

PR #555, opened 2016-03-10
[LRN] add \mathbb{} font with support for those letters: C, H, N, P, …

karoltarasiuk:
…Q, R, Z

Conflicts:
	src/commands/math/commands.js
	src/css/math.less
	test/unit/latex.test.js
	test/unit/typing.test.js

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -70,12 +70,80 @@ var Style = P(MathCommand, function(_, super_) {
   };
 });
 
+var DoubleStruck = P(Variable, function(_, super_) {
+  _.symbols = {
+    C: "&#8450;",
+    H: "&#8461;",
+    N: "&#8469;",
+    P: "&#8473;",
+    Q: "&#8474;",
+    R: "&#8477;",
+    Z: "&#8484;"
+  };
+  _.init = function(ch) {
+    var inner = ch;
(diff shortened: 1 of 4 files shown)


laughinghan:
Since it's not obvious what The Right Thing is to do here, can you spec out a bit what the intended behavior of this code is? For example, it's not clear from your tests how letters not currently supported as doublestruck in our font (ABCXYZ) are intended to appear.

Tangent:
I was thinking recently that there's no reason to be constrained to the model where, just because a command takes arguments in LaTeX, it must be a block in MathQuill. What if instead, after you typed `\mathbb`, it prompted you to type one more letter&mdash;and it had to be one of CHNPQRZ, it would refuse to let you proceed unless you typed one of those or hit Backspace/Delete/Escape&mdash;and then it inserted the corresponding doublestruck letter, which would thereafter just be a symbol? If you typed it wrong, if you typed `\mathbb Q` and got ℚ when you meant to get ℝ, well, you're not in any special mode, you just backspace the ℚ and type `\mathbb R` to get ℝ. After all, when do you ever actually want to type `\mathbb{CHNPQRZ}`? You'd be constrained to only being able to parse `\mathbb{Q}\mathbb{R}` etc ofc, parsing `\mathbb{QR}` or `\mathbb{A}` would be invalid.

PR #556, opened 2016-03-10
Feature/underset

karoltarasiuk:


stufflebear:
Verified latex commands: http://mathb.in/53603

Looks good too: 
screenshot 2016-03-10 17 12 13

stufflebear:
I just noticed that the spacing on the fraction changed. Was that intentional?

src/commands/math/commands.js
```diff
@@ -429,6 +429,36 @@ CharCmds['/'] = P(Fraction, function(_, super_) {
...
+
+LatexCmds.overset = P(MathCommand, function(_, super_) {
+  _.ctrlSeq = '\\overset';
+  _.htmlTemplate =
+      '<span class="mq-overset mq-overunder mq-non-leaf">'
+    +   '<span class="mq-over">&0</span>'
+    +   '<span class="mq-under">&1</span>'
+    + '</span>'
+  ;
+  _.textTemplate = ['[', '|', ']'];
```
laughinghan: there's no need to change this because we don't really care about .text(), but out of curiosity was there any particular reason you chose this?

src/css/math.less
```diff
@@ -241,14 +241,45 @@
...
+    position: relative;
+    top: .01em;
+
+    // fix \mathbf vertical alignment issue
+    .mq-font {
+      font-family: inherit;
+    }
+
+    &.mq-underset {
+      vertical-align: text-top;
```
laughinghan: this doesn't quite work, it will mis-align the underset if the contents are taller than normal text:
image
For the same LaTeX, `1 + \underset{ abc }{ \frac{2}{3} } + 2`, MathJax renders:
image


I think you'll need the `float` and `width: 100%` trick that fraction uses to do the right thing to the baseline.

src/css/math.less
```diff
@@ -241,14 +241,45 @@
...
+  // \overset & \underset
+  // overset and underset text
+  .mq-overunder {
+    text-align: center;
+    position: relative;
+    top: .01em;
+
+    // fix \mathbf vertical alignment issue
+    .mq-font {
+      font-family: inherit;
```
laughinghan: this causes Symbola to be used, right? There's no proper bold and italics and other styling for Symbola; what was the vertical alignment issue? Is it also fixed by fixing https://github.com/mathquill/mathquill/pull/556/files#r55783521 ?

nadavkav:
Thank you for the patch!
We use mathquill in Moodle shortmath question type:
https://moodle.org/plugins/qtype_shortmath
And we needed to add the following CSS to make it fully functional:
```

.mq-math-mode .mq-underset {
  font-size: 70%;
  text-align: center;
  vertical-align: 0.13em;
  padding: 0 0.2em;
  display: -moz-groupbox;
  display: inline-block;
}

.mq-math-mode .mq-overset {
  font-size: 70%;
  text-align: center;
  vertical-align: 0.92em;
  padding: 0 0.2em;
  display: -moz-groupbox;
  display: inline-block;
}

.mq-math-mode .mq-overset .mq-over,
.mq-math-mode .mq-underset .mq-over {
  display: block;
  padding: 0 0.1em;
}

.mq-math-mode .mq-overset .mq-under,
.mq-math-mode .mq-underset .mq-under {
  display: block;
  /*border-top: 1px solid;*/
  float: right;
  width: 100%;
  margin-top: -0.21em;
  /*padding: 0.1em;*/
}
```

PR #557, opened 2016-03-10
[LRN] Add chemical bond symbols from mhchem

karoltarasiuk:
These are standalone symbols - only the bond can be wrapped in \ce as
nothing else will be recognised.

\ce{\bond{-}}
\ce{\bond{=}}
\ce{\bond{#}}
\ce{\bond{...}}
\ce{\bond{....}}
\ce{\bond{->}}
\ce{\bond{<-}}

Conflicts:
	src/css/math.less
	test/unit/typing.test.js

### DIFF
--- a/src/commands/text.js
+++ b/src/commands/text.js
@@ -317,6 +317,84 @@ LatexCmds.uppercase =
 LatexCmds.lowercase =
   makeTextBlock('\\lowercase', 'span', 'style="text-transform:lowercase" class="mq-text-mode"');
 
+// Chemical bonds in mhchem format
+LatexCmds.ce = P(TextBlock, function(_, super_) {
+  function wrapDots(html) {
+    return html.replace(/(&middot;)/g, '<span class="mq-bond-dot">$1</span>');
+  }
+  _.bondHtml = {
+    '-': '-',
+    '=': '=',
+    '#': '&#8801;',
+    '...': wrapDots('&middot;&middot;&middot;'),
+    '....': wrapDots('&middot;&middot;&middot;&middot;'),
+    '->': '&#8594;',
+    '<-': '&#8592;',
+  };
+  _.bond = null;
+  _.ctrlSeq = '\\ce';
+  _.regex = /^(\\bond{([^{}]+?)})/i;
+  _.jQadd = function(jQ) {
+    super_.jQadd.call(this, jQ);
+    if (this.bond) this.jQaddBond();
+  };
+  _.jQaddBond = function() {
+    this.jQ.empty()
+      .removeClass('mq-text-mode')
+      .addClass('mq-ce mq-bond')
+      .html(this.bondHtml[this.bond]);
+  };
+  _.parseBond = function (contents) {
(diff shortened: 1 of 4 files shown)


test/unit/typing.test.js
```diff
@@ -1046,4 +1046,39 @@ suite('typing with auto-replaces', function() {
...
   });
+
+  suite('Chemistry Symbols', function() {
+    test('add bond via mq.write', function() {
+      mq.write('\\ce{\\bond{#}}');
+      assertLatex('\\ce{\\bond{#}}');
+    });
+    test('invalid bond types not added', function() {
+      mq.write('\\ce{\\bond{x}}');
+      assertLatex('\\ce{}');
```
laughinghan: if the bond type is invalid shouldn't it be a parse failure? Don't we want `assertLatex('')`?

src/commands/text.js
```diff
@@ -317,6 +317,84 @@ LatexCmds.uppercase =
 LatexCmds.lowercase =
   makeTextBlock('\\lowercase', 'span', 'style="text-transform:lowercase" class="mq-text-mode"');
 
+// Chemical bonds in mhchem format
+LatexCmds.ce = P(TextBlock, function(_, super_) {
+  function wrapDots(html) {
+    return html.replace(/(&middot;)/g, '<span class="mq-bond-dot">$1</span>');
```
laughinghan: any reasons for the parens in the regex?

src/commands/text.js
```diff
@@ -317,6 +317,84 @@ LatexCmds.uppercase =
...
+  _.bond = null;
+  _.ctrlSeq = '\\ce';
+  _.regex = /^(\\bond{([^{}]+?)})/i;
+  _.jQadd = function(jQ) {
+    super_.jQadd.call(this, jQ);
+    if (this.bond) this.jQaddBond();
+  };
+  _.jQaddBond = function() {
+    this.jQ.empty()
+      .removeClass('mq-text-mode')
```
laughinghan: why was it ever `.mq-text-mode`?

test/unit/typing.test.js
```diff
@@ -1046,4 +1046,39 @@ suite('typing with auto-replaces', function() {
...
+    });
+    test('multiple bonds can be written', function() {
+      mq.write('\\ce{\\bond{....}}\\ce{\\bond{<-}}');
+      assertLatex('\\ce{\\bond{....}}\\ce{\\bond{<-}}');
+    });
+    test('cursor seeks to just outside bond symbol', function() {
+      mq.write('\\ce{\\bond{#}}\\sigma');
+      $(mq.el()).find('.mq-ce').trigger('mousedown');
+      mq.typedText('\\pi');
+      assertLatex('\\ce{\\bond{#}}\\pi\\sigma');
```
laughinghan: huh, why does this work?

laughinghan:
This looks super useful! Why'd you inherit from `TextBlock`s, though? Since you're trying to produce bond symbols, any reason not to inherit from `Symbol`?

PR #561, opened 2016-03-10
[LRN] Added recurring decimal dot symbol

karoltarasiuk:
Conflicts:
    src/commands/math/commands.js
    src/css/math.less

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -81,6 +81,15 @@ LatexCmds.underline = bind(Style, '\\underline', 'span', 'class="mq-non-leaf mq-
 LatexCmds.overline = LatexCmds.bar = bind(Style, '\\overline', 'span', 'class="mq-non-leaf mq-overline"');
 LatexCmds.overrightarrow = bind(Style, '\\overrightarrow', 'span', 'class="mq-non-leaf mq-overarrow mq-arrow-right"');
 LatexCmds.overleftarrow = bind(Style, '\\overleftarrow', 'span', 'class="mq-non-leaf mq-overarrow mq-arrow-left"');
+LatexCmds.dot = P(MathCommand, function(_, super_) {
+    _.init = function() {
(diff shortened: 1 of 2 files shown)


src/commands/math/commands.js
```diff
@@ -79,6 +90,7 @@ LatexCmds.mathtt = bind(Style, '\\mathtt', 'span', 'class="mq-monospace mq-font"
 //text-decoration
 LatexCmds.underline = bind(Style, '\\underline', 'span', 'class="mq-non-leaf mq-underline"');
 LatexCmds.overline = LatexCmds.bar = bind(Style, '\\overline', 'span', 'class="mq-non-leaf mq-overline"');
+LatexCmds.dot = bind(DotRecurringDecimalStyle, '\\dot');
```
laughinghan: the only reason to use `bind()` is for reusable, "abstract" classes like `VanillaSymbol` or `Style`. For classes that correspond to just one command (which is actually most non-symbols), I'd prefer inlining it all, like for example [`TextColor`](https://github.com/mathquill/mathquill/pull/561/files#diff-4a4c67b9876a02660f57ba183147e639R102). In this case that might be something like:
```js
LatexCmds.dot = P(MathCommand, function(_, super_) {
  _.init = function() {
    super_.init.call(this, '\\dot', ...
```

karoltarasiuk:
@laughinghan @stufflebear Suggested changes implemented

laughinghan:
Beautiful!

xutaogit:
@karoltarasiuk  excuse me, how can I show recurring decimal dot in MathQuill.
It seems can not match the right latex string,thanks.
image

karoltarasiuk:
Hmmm, seems the symbol is still there: https://github.com/mathquill/mathquill/blame/master/src/commands/math/commands.js#L86

But when typing it gets translated to the `\text{dot}` command...
Screen Shot 2019-03-12 at 11 38 14 am

It's interesting because other commands, e.g. `\underline` seem to work just fine.

@laughinghan any idea? I forgot and not using MathQuill anymore, but it was supposed to work...

PR #572, opened 2016-03-14
Fix #571

### LINKED ISSUE #571: Spaces are recognized as Variables
If I input a space in an expression followed by a variable (e.g. `1+ a` instead of `1+a`), `text()` returns it as `1+ *a`, instead of `1+a`.

The-Penultimate-Defenestrator:
No more rogue multiplication signs.

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -26,7 +26,7 @@ var Variable = P(Symbol, function(_, super_) {
     var text = this.ctrlSeq;
     if (this[L] && !(this[L] instanceof Variable)
         && !(this[L] instanceof BinaryOperator)
-        && this[L].ctrlSeq != "\\ ")
+        && this[L].ctrlSeq !== "\\ ")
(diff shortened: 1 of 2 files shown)


src/commands/math/basicSymbols.js
```diff
@@ -25,7 +25,8 @@ var Variable = P(Symbol, function(_, super_) {
   _.text = function() {
     var text = this.ctrlSeq;
     if (this[L] && !(this[L] instanceof Variable)
-        && !(this[L] instanceof BinaryOperator))
+        && !(this[L] instanceof BinaryOperator)
+        && this[L].ctrlSeq != "\\ ")
```
kschaefe: Is there a reason not to use `!==`?

laughinghan:
Love it

The-Penultimate-Defenestrator:
It's small but fixes something. I hope to contribute more until this project reaches the point where I can use it in my app. Which just means fixing one or two more bugs in `text()`

PR #574, opened 2016-03-14
Fix variables raised to exponents

The-Penultimate-Defenestrator:
Before, variables raised to exponents would insert three stars (`x***2`) instead of two as intended (`x**2`). It looks like someone tried to implement this before (or did, and it broke), but now it works as intended. I also added a test case in case it breaks again in the future.

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -29,7 +29,7 @@ var Variable = P(Symbol, function(_, super_) {
         && this[L].ctrlSeq !== "\\ ")
       text = '*' + text;
     if (this[R] && !(this[R] instanceof BinaryOperator)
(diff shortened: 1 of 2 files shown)


stufflebear:
Thanks for the fix! We recently changed .text to output `^` instead of `**` in https://github.com/mathquill/mathquill/pull/526. So a variable with an exponent  currently returns `x*^2` on master. Can you rebase this off of the latest master and update the test?

The-Penultimate-Defenestrator:
Sure. I had some minor problems with building before, but I'll try again in
a few minutes when I get home from school.

The-Penultimate-Defenestrator:
Done!

laughinghan:
Awesome, thanks so much!!

PR #577, opened 2016-03-15
[LRN] Change \triangle and \ddots character bindings

karoltarasiuk:
\triangle is now a bigger triangle
\ddots now points down and right


karoltarasiuk:
@laughinghan turned out `\ddots` were fixed by you anyway, for consistency with e.g. MathJax. Only `\triangle` then

laughinghan:
Before: small triangle

After: big triangle

MathJax ([via MathBin](http://mathb.in/53852)): MathJax

A marked improvement!

Yeah `\ddots` has been fixed on mainline but not by me, that was thanks to @robinwassen and @deannah :smiley:

PR #578, opened 2016-03-16
Added .text method for the command '*'

stufflebear:
The latex for this is \cdot, which makes less sense for the text output. This change makes typed input using '*' compatible with tools like mathjs.

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -416,14 +416,8 @@ LatexCmds['±'] = LatexCmds.pm = LatexCmds.plusmn = LatexCmds.plusminus =
 LatexCmds.mp = LatexCmds.mnplus = LatexCmds.minusplus =
   bind(PlusMinus,'\\mp ','&#8723;');
 
-var Times = P(BinaryOperator, function(_) {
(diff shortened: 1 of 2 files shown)


src/commands/math/basicSymbols.js
```diff
@@ -416,8 +416,14 @@ LatexCmds['±'] = LatexCmds.pm = LatexCmds.plusmn = LatexCmds.plusminus =
...
 
+var Times = P(BinaryOperator, function(_) {
+  _.text = function() {
+    return '*'
+  };
+});
+
 CharCmds['*'] = LatexCmds.sdot = LatexCmds.cdot =
-  bind(BinaryOperator, '\\cdot ', '&middot;');
+  bind(Times, '\\cdot ', '&middot;');
```
laughinghan: It looks like `BinaryOperator` can just take in/be bound a 3rd argument that'll be the text, like https://github.com/mathquill/mathquill/pull/578/files#diff-50c9fc0c437e0937042e370a2e965768R482 , does that work here?

stufflebear: Yeah, that works for this

Issue #559, opened 2016-03-10
Keyboard input not working in iOS 7


drlippman:
OK, I know iOS 7.0.3 is way old, so maybe it's not worth worrying about, but for whatever reason the on-screen keyboard doesn't seem to be able to input into MathQuill.  I tried it on the mathquill.com page input box, and typing produced no characters, both in Safari and Chrome.

I checked, and keydown and keypress events are being fired in textareas, so not sure why it would be having problems.

drlippman:
I'm having issues with other functions too, like API calls, so maybe there's a larger incompatibility. Unfortunately I don't have the necessary setup to debug properly.

drlippman:
Interesting. I'm having similar problems using the onscreen keyboard to type in the box at mathquill.com using my Android Nexus 5x running Chrome.

laughinghan:
I mean it's worrying to me...

laughinghan:
which version of Android Chrome? Think you could get the User Agent from something like http://whatsmyuseragent.com/ ?

drlippman:
Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29Q) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.95 Mobile Safari/537.36
On Mar 9, 2016 8:58 PM, Han Seoul-Oh <notifications@github.com> wrote:which version of Android Chrome? Think you could get the User Agent from something like http://whatsmyuseragent.com/ ?

drlippman:
I updated my iPad to iOS 9 and typing using the onscreen keyboard is working fine now, so it must have been an issue in iOS 7.  Still having the issue on my Andriod.

The-Penultimate-Defenestrator:
Text selection still does not work on iOS 9. (used for things like selecting text and pressing `/` for putting things in the numerator)

laughinghan:
[...]

PR #585, opened 2016-03-18
Kill `font-size: 0`, fixes many typing issues

### LINKED ISSUE #540: Keyboard Input not working with Chrome Canary v51.0.2669.0
I realize Chrome Canary is not a stable browser, nor always worth supporting.  On the other hand, being able to type into Mathquill fields is pretty central to the experience, so I thought I would flag it.

Keyboard input events of any sort do not seem to be triggering any effect in today's (7 Mar, 2016) Chrome Canary build (51.0.2669.0).  This happens in Windows 10 and OS X Yosemite.

### LINKED ISSUE #566: Pressing Enter doesn't trigger 'enter' handler
To repro:
1. on `demo.html`, go to the Console and do `MQ($('#editable-math')[0]).config({handlers: {enter: function() { console.log('enter'); }}})`
2. go into the math field and hit Enter

Expected: `enter` is logged
Observed: nope

### LINKED ISSUE #584: Hide iOS blinking blue cursor *without* breaking typing in Chrome
So maybe setting `font-size: 0` on the textarea to hide the blinking blue cursor on iOS was not so ingenious after all (323724b20c5a18eec1147d23b2cc65f13686f930 on #452), it's causing:
- all keyboard input to be ignored in Chrome Canary 51 (#540, which I've repro'd and confirmed goes away without the `font-size: 0`, and reported to [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=595943)),
- the Enter key to be ignored in Chrome 48 and Safari 9 (MathQuill doesn't let you type newlines but is supposed to fire an `enter` handler: #566; reported to [Webkit](https://bugs.webkit.org/show_bug.cgi?id=155615#c0) and mentioned in the Chromium issue),
- in Safari on iOS 9.2.1 (and probably other mobile browsers too), focusing a math field zooms _wayyyy_ in (happens on the http://mathquill.com homepage and local test page /tests/demo.html),
- and probably the other typing issues as well (#559).

Okay so `font-size: 0` is not going to cut it. In that case, how do we hide the blinking blue cursor in iOS? Options:

- position the hidden textarea past the left (or top) edge of the viewport. This is a perennially popular solution, it's been suggested in this project a few times (9b323e3347ea4be55fd77e75f17a37d08889f9dc, 07578edd7537f1000005b55f0b143acc696223c3) and is the only suggestion that works for iOS and IE (and not just Gecko and Webkit like used to be the case for MathQuill's hidden textarea (1fa701b84b0e53e46bebb9628ad1ceeb04b08fc3)) that I've seen on StackOverflow (e.g. [this elegantly hand-drawn diagram](http://stackoverflow.com/a/30010698/362030), or [here](http://stackoverflow.com/a/29146606/362030) or [here](http://stackoverflow.com/a/30950258/362030))
    + Since this is clearly what most of the Internet is doing, it's a good idea to have this in our back pocket.
    + However, this technique has a major caveat: when you focus a textarea, all browsers scroll it into view, and mobile browsers zoom into it. This doesn't matter to full-page webapps that can't scroll at all like the Desmos graphing calculator, but if your page or app can scroll and our hidden textarea is past the top edge of the viewport, we'll scroll to the top whenever MathQuill gets focused. Which is why many implementations of this technique put the hidden textarea past the left edge of the viewport without affecting it's vertical coordinate, this way desktop browsers usually scroll to exactly the right place, as do mobile browsers on pages with [`user-scalable=no`](https://developer.apple.com/library/iad/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW6). However, this is still problematic on normal pages in mobile browsers, because this means that when focusing a MathQuill math field, they'll zoom into the left edge of the screen even if the math field is in the center of the screen (similarly, on a page wide enough to need horizontal scrolling, desktop browsers will scroll all the way to the left trying to scroll-into-view the textarea; but pages needing horizontal scrolling in a desktop browser are pretty uncommon).
- browser-sniff for iOS (or mobile?) and selectively apply `line-height: 0` to the textarea (which has the same desired behavior of hiding the blinking blue cursor on iOS and undesired behavior of breaking typing sometimes; but doesn't have the issue where Safari on iOS zooms _wayyyy_ in when focused)
    + This feels like admitting defeat. We avoided browser sniffing when none of the straightforward ways to vertically align the fraction bar correctly always failed in at least one of Gecko, Webkit, and Opera; we avoided browser sniffing when keyboard events were ["madness"](http://unixpapa.com/js/key.html); we're gonna be defeated by a goddamn blinking cursor?
- I did find one new technique that I hadn't discovered during #452, injecting a newline at the beginning of a textarea that's less than one line and scrolling to the top, so that the cursor is always out of view: http://jsbin.com/cohema/edit?css,js,output
    + I really do appear to _have to_ inject a newline, [my other attempts](https://gist.github.com/laughinghan/faba0e276a5c3ea30088) to insert space and scroll the cursor out of view all result in the textarea increasing in height/width until a bit of the blinking blue cursor "peeks out".
    + Unfortunately injecting a newline means this probably isn't a viable near-term solution, since that interferes with us figuring out what people typed by [reading the contents of the textarea](https://github.com/mathquill/mathquill/blob/v0.10.0/src/services/saneKeyboardEvents.util.js#L191) (that's the whole point of the hidden textarea in the first place); we would have to change our already disgustingly complex keyboard events cross-browser madness-queller to ignore the injected newline, without ignoring a real newline that gets typed (#566 is what started me on this in the first place).

Out of all of these, the best option looks like browser-sniffing :disappointed: . Unlike the inject-newline-and-scroll-down thing, it's plug-and-play, we don't have to futz with the keyboard events code; and it has a theoretical potential future downside, but the positioning the textarea past the edge of the viewport has a concrete downside today.

laughinghan:
Resolves #584 "Hide iOS blinking blue cursor *without* breaking typing in Chrome".

In Safari on my iPhone 5 running iOS 9.2.1, the blinking blue cursor
doesn't show up at all when focusing and typing in this thing:
    http://jsbin.com/sodaji/edit?html,css,js,output
And in all browsers typing in that works just fine, including typing
newlines by pressing Enter or combos like Shift-Enter.

And with this patch, typing on `/test/demo.html` works great on my iPhone:
    screenshot
(Notice that it's not zoomed wayyyy in, like it used to)

As detailed by #584:
Fixes #540
Fixes #566
Hopefully addresses #559?

### DIFF
--- a/src/css/mixins/css3.less
+++ b/src/css/mixins/css3.less
@@ -5,6 +5,13 @@
   -o-transform-origin: @arguments;
   transform-origin: @arguments;
 }
+.transform (...) {
+  -webkit-transform: @arguments;
+  -moz-transform: @arguments;
+  -ms-transform: @arguments;
+  -o-transform: @arguments;
+  transform: @arguments;
+}
 
 .user-select (...) {
   -webkit-user-select: @arguments;
--- a/src/css/textarea.less
+++ b/src/css/textarea.less
@@ -14,7 +14,7 @@
     position: absolute; // the only way to hide the textarea *and* the
     clip: rect(1em 1em 1em 1em); // blinking insertion point in IE
 
-    font-size: 0; // the only way to hide the blinking blue cursor in iOS 8
+    .transform(scale(0)); // the only way to hide the blinking blue cursor in iOS 8 #584
 
     resize: none; // hotfix: https://code.google.com/p/chromium/issues/detail?id=355199#c1
 
--- a/src/services/keystroke.js
+++ b/src/services/keystroke.js
@@ -129,9 +129,6 @@ Node.open(function(_) {
       while (cursor[L]) ctrlr.selectLeft();
       break;
 
-    case 'Enter':
-      return ctrlr.handle('enter');
-
     default:
       return;
     }
--- a/src/services/textarea.js
+++ b/src/services/textarea.js
@@ -86,6 +86,7 @@ Controller.open(function(_) {
     this.focusBlurEvents();
   };
   _.typedText = function(ch) {
+    if (ch === '\n') return this.handle('enter');
     var cursor = this.notify().cursor;
     cursor.parent.write(cursor, ch);
     this.scrollHoriz();


stufflebear:
:shipit:

PR #594, opened 2016-03-23
Move integral from advancedSymbols to commands.

jwmerrill:
This is where sums and products are, and I think it makes sense to keep them together.


laughinghan:
hmm, it _is_ just a symbol, is the funny thing with moving it out of `advancedSymbols.js` (unless it should gonna start coming with lower/upper limits too? Although I guess being in a `<big>` is different too). Do you need it in MathQuill-Basic or something?

jwmerrill:
> Do you need it in MathQuill-Basic or something?

Yes.

laughinghan:
That's fine I guess +1

jwmerrill:
I'd be happy to move it into basicSymbols instead if you prefer.

laughinghan:
Don't really care, just want to record why

Issue #530, opened 2016-02-25
Distorted Symbola font rendering in IE for some font sizes


karoltarasiuk:
I'm attaching 2 screenshots with the issue. I created a page (full HTML source below) generating a digit of 6 in different sizes, from 10px to 40px.

Chrome (no distortion):
![screen shot 2016-02-25 at 2 50 43 pm](https://cloud.githubusercontent.com/assets/1845230/13309282/853771d0-dbcf-11e5-9dae-c4a32df025cd.png)

IE Edge (visible distortion, especially for lower font sizes, worst is probably for 14px and 16px):
![screen shot 2016-02-25 at 2 50 30 pm](https://cloud.githubusercontent.com/assets/1845230/13309305/b79c7198-dbcf-11e5-9e20-57e00152906c.png)

The issue is also visible on IE10 & 11. From what I see those browsers load `Symbola.woff` file. We also observed this issue with EOT and TTF within our product, so it seems like it's a general issue with this font in IE.

Source (based on your `demo.html` page I've built this one):
```
<!DOCTYPE html>
<html>
<head>
<title>MathQuill - Font issue in IE</title>
<link rel="stylesheet" type="text/css" href="support/home.css">
<link rel="stylesheet" type="text/css" href="../build/mathquill.css">

<style type="text/css">
code span {
  font: 90% Verdana, sans-serif;
}
#codecogsimg {
  vertical-align: middle;
  border: none;
}
.mathquill-text-field {
  width: 100%;
}
#html-source {
  display: none;
  font-size: 90%;
  white-space: pre-wrap;
}
.mq-math-mode .mq-editable-field {
  min-width: 1cm;
}
</style>

</head>
<body>
    <div id="body">
        <p>Font issue in IE 10, 11 and Edge (possibly older too):</p>
        <p class="sixes"></p>
    </div>
<script type="text/javascript" src="support/jquery-1.7.2.js"></script>
<script type="text/javascript" src="../build/mathquill.js"></script>
<script type="text/javascript">
MQ = MathQuill.getInterface(MathQuill.getInterface.MAX);
$(function() {
    var i,
        sixes = '',
        start = 10,
        end = 40;
    for (i = start; i <= end; i++) {
        sixes += '<span style="font-size: ' + i + 'px" class="mathquill-math-field">6<\/span>';
    }
    $('.sixes').html(sixes);
    $('.mathquill-math-field').each(function() { MQ.MathField(this); });
});
</script>
</body>
</html>

```

[...]

Issue #324, opened 2014-06-16
Stylesheets of test pages don't work in IE≥9


laughinghan:
Something something IE9 ignores stylesheets unless they're `text/css` (our server doesn't set `Content-Type` and Node apparently defaults everything to `text/plain`): http://msdn.microsoft.com/en-us/library/ie/gg622939(v=vs.85).aspx

Local fix:

``` diff
diff --git a/script/test_server.js b/script/test_server.js
index bf7d84d..51e3e2a 100644
--- a/script/test_server.js
+++ b/script/test_server.js
@@ -35,6 +35,8 @@ function serveRequest(req, res) {
         }
       }
       else {
+        var ext = filepath.match(/\.[^.]+$/);
+        if (ext) res.setHeader('Content-Type', 'text/' + ext[0].slice(1));
         res.end(data);
       }

```

Should fix properly.

PR #598, opened 2016-03-31
[LRN] Added WOFF and WOFF2 fonts built from OTF

karoltarasiuk:


laughinghan:
Sweet! What tool did you use to build the WOFF{,2} files? Also, they're used by non-IE too right, let's double-check those are fine?

karoltarasiuk:
I've used this one: https://font-converter.net/en

I've uploaded OTF and generated WOFF and WOFF2 only. I checked our demo page on Edge, Chrome, Firefox and Safari and looks OK.

laughinghan:
Let's add the test case from #530 to our visual tests: https://github.com/Learnosity/mathquill/pull/49

Mind posting screenshots? I'll take care of OS X (10.10.5):

Chrome 48:
image

Firefox 45:
image

Safari 9.0.3:
image

The scaled-down 5 doesn't look great, but overall there's no glaring problems like there were in IE.

karoltarasiuk:
Edge on Win10:
screen shot 2016-04-01 at 9 51 57 am
Chrome on Win10:
screen shot 2016-04-01 at 9 57 39 am
FF on Win10:
screen shot 2016-04-01 at 9 58 21 am
Safari 5 on Win10 (you can't download newer for Windows from anywhere):
screen shot 2016-04-01 at 10 01 22 am
IE11 on Win8 (error thrown in the console: math field 337px wide instead of 102px):
screen shot 2016-04-01 at 10 07 39 am
Same on IE10 on Win7, and IE9 on Win7
IE8 on Win7:
screen shot 2016-04-01 at 10 09 57 am

Red pages in IEs above are not connected to the font change though, so I don't think it's a blocker for this PR.

laughinghan:
Hahaha IE8 and Edge work but not IE9-11? It looks like the stylesheet didn't load, probably #324

test/visual.html
```diff
@@ -197,6 +197,20 @@
...
 <tr><td><span class="mathquill-static-math">\sum _{n=0}^3\cos x</span><td><span>\sum _{n=0}^3\cos x</span>^M
+<tr><td colspan=2><span id="sixes"></span>
+<script>
+$(function() {
+    var i,
+        sixes = '',
+        start = 10,
+        end = 40;
+    for (i = start; i <= end; i++) {
+        sixes += '<span style="font-size: ' + i + 'px" class="mathquill-math-field">6<\/span>';
```
laughinghan: @karoltarasiuk: just noticed the backslash in `<\/span>`. Fun fact: this was specified to be necessary in HTML4 via SGML, but was never actually necessary in any browser and is unnecessary in HTML5/the HTML Living Standard (summary: http://stackoverflow.com/a/236106/362030 , long version: http://mths.be/etago )

PR #603, opened 2016-04-06
Feature: integrals' from/to now mandatory, prettier

laughinghan:
Depends on the tweak to CSS styles for `big` from #602.

Like #258 but for integrals rather than sums and products, integrals now
come with from and to blocks as an inextricable part of them.

Note that this means that integrals with only a subscript, e.g.
Stoke's theorem `\int_{\partal\Omega} \omega = \int_\Omega d\omega`,
can't be typed anymore, but #258 made an analogous sacrifice.

I originally intended this to be two PRs, one to fix how integrals look
and one to make the from/to blocks mandatory, but was stymied by the
vertical alignment of the superscript block (see the second test case).
The only way I could figure out to get that to work was .mq-sup-inner,
which doesn't exist in normal sup/sub blocks, I'm only able to add it by
making from/to blocks mandatory and inserting a span.mq-sup-inner into
the HTML template (luckily mandatory from/to blocks are what Desmos
needs anyway).

One nice thing: supsub no longer has to autodetect whether it's
currently serving as the from/to of an integral, we can get rid of all
the .mq-limit stuff AND the SupSub::respace() stuff, and get rid of it
from the redrawing/reflowing test.

Speaking of tests, there wasn't enough space for the spherical integral
in the rendered math column, so I made it wider, half instead of a third
of the page.

----
Before:
before

After:
after

KaTeX:
katex

[MathJax](http://mathb.in/55065):
mathjax

LaTeXiT:
katex

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -181,7 +181,6 @@ var SupSub = P(MathCommand, function(_, super_) {
         break;
       }
     }
-    this.respace();
   };
   Options.p.charsThatBreakOutOfSupSub = '';
   _.finalizeTree = function() {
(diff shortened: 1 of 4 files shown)


jwmerrill:
@laughinghan did this actually make it onto master? I see that it was merged into fix.visual-rendering, but I'm not seeing it on master.

jwmerrill:
Looks like it did not make it onto master. Opened https://github.com/mathquill/mathquill/pull/627 to actually get this on master.

PR #604, opened 2016-04-06
Fix auto-commands eating letters from operator names

laughinghan:
Should have lower "precedence".

Auto-commands used to be the first thing that `Letter::createLeftOf()`
checked for. Now instead it first waits for the dust to settle on
auto-operator names, then seeks nearby letters but stops at the edge of
an operator name. Also note that before, this letter itself had not been
inserted yet; now it has so the seeking and the fragment that's removed
both start from the this letter.

### DIFF
--- a/test/unit/typing.test.js
+++ b/test/unit/typing.test.js
@@ -813,6 +813,7 @@ suite('typing with auto-replaces', function() {
   suite('autoCommands', function() {
     setup(function() {
       MQ.config({
+        autoOperatorNames: 'sin pp',
         autoCommands: 'pi tau phi theta Gamma sum prod sqrt nthroot'
       });
     });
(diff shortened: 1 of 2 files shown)


test/unit/typing.test.js
```diff
@@ -880,6 +881,13 @@ suite('typing with auto-replaces', function() {
       assertLatex('\\sin\\pi');
     });
 
+    test('has lower "precedence" than operator names', function() {
+      mq.typedText('ppi');
+      assertLatex('\\operatorname{pp}i');
+      mq.keystroke('Left Left').typedText('i');
+      assertLatex('\\pi pi');
```
jwmerrill: Wait, why is this not now "\\pi \\pi"?

jwmerrill:
Otherwise, looks good to me.

Although I must say that the code being changed here is not the most, um, self documenting...

test/unit/typing.test.js
laughinghan: because auto-conversion only happen when you type their last letter, nearby words aren't auto-converted. If you typed `i`, then moved the cursor in front and typed a `p`, you'd get `pi` not `π`, and nothing typed near it should auto-convert it (except backspacing and re-typing the `i`)

laughinghan:
> Although I must say that the code being changed here is not the most, um, self documenting...

Yeah. :disappointed: Ideas? More comments?

PR #607, opened 2016-04-08
Minus after comma is unary negation

### LINKED ISSUE #523: Minus and Negative should get different amounts of trailing space
image

image

arnog:
Fix #523: If the left sibling of a PlusMinus operator is a punctuation, consider the operator to be unary. Fix the case where "1,-1-1" would consider the first "-" to be a binary operator.

Here are the before and after screenshots...
screenshot 2016-04-08 15 01 52
screenshot 2016-04-08 15 04 21

### DIFF
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -321,5 +321,6 @@ LatexCmds.cap = LatexCmds.intersect = LatexCmds.intersection =
 
 // FIXME: the correct LaTeX would be ^\circ but we can't parse that
 LatexCmds.deg = LatexCmds.degree = bind(VanillaSymbol,'\\degree ','&deg;');
+
 LatexCmds.ang = LatexCmds.angle = bind(VanillaSymbol,'\\angle ','&ang;');
 LatexCmds.measuredangle = bind(VanillaSymbol,'\\measuredangle ','&#8737;');
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -220,10 +220,11 @@ LatexCmds.opencurlybrace = LatexCmds.lbrace = bind(VanillaSymbol, '\\lbrace ', '
 LatexCmds.closecurlybrace = LatexCmds.rbrace = bind(VanillaSymbol, '\\rbrace ', '}');
 LatexCmds.lbrack = bind(VanillaSymbol, '[');
 LatexCmds.rbrack = bind(VanillaSymbol, ']');
-
+ 
 //various symbols
 LatexCmds.slash = bind(VanillaSymbol, '/');
 LatexCmds.vert = bind(VanillaSymbol,'|');
+LatexCmds['|'] = bind(VanillaSymbol,'\\| ','&#8741;');
 LatexCmds.perp = LatexCmds.perpendicular = bind(VanillaSymbol,'\\perp ','&perp;');
 LatexCmds.nabla = LatexCmds.del = bind(VanillaSymbol,'\\nabla ','&nabla;');
 LatexCmds.hbar = bind(VanillaSymbol,'\\hbar ','&#8463;');
@@ -321,6 +322,12 @@ LatexCmds.cap = LatexCmds.intersect = LatexCmds.intersection =
 
 // FIXME: the correct LaTeX would be ^\circ but we can't parse that
 LatexCmds.deg = LatexCmds.degree = bind(VanillaSymbol,'\\degree ','&deg;');
+
+LatexCmds.celsius = bind(VanillaSymbol,'\\celsius ','&#8451;');
+LatexCmds.micro = bind(VanillaSymbol,'\\micro ','&#181;');
+LatexCmds.perthousand = bind(VanillaSymbol,'\\perthousand ','&#8240;');
+LatexCmds.ohm = bind(VanillaSymbol,'\\ohm ','&#8486;');
+
 
 LatexCmds.ang = LatexCmds.angle = bind(VanillaSymbol,'\\angle ','&ang;');
 LatexCmds.measuredangle = bind(VanillaSymbol,'\\measuredangle ','&#8737;');
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -220,11 +220,10 @@ LatexCmds.opencurlybrace = LatexCmds.lbrace = bind(VanillaSymbol, '\\lbrace ', '
 LatexCmds.closecurlybrace = LatexCmds.rbrace = bind(VanillaSymbol, '\\rbrace ', '}');
 LatexCmds.lbrack = bind(VanillaSymbol, '[');
 LatexCmds.rbrack = bind(VanillaSymbol, ']');
- 
+
 //various symbols
 LatexCmds.slash = bind(VanillaSymbol, '/');
 LatexCmds.vert = bind(VanillaSymbol,'|');
-LatexCmds['|'] = bind(VanillaSymbol,'\\| ','&#8741;');
 LatexCmds.perp = LatexCmds.perpendicular = bind(VanillaSymbol,'\\perp ','&perp;');
 LatexCmds.nabla = LatexCmds.del = bind(VanillaSymbol,'\\nabla ','&nabla;');
 LatexCmds.hbar = bind(VanillaSymbol,'\\hbar ','&#8463;');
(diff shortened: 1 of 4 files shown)


src/services/latex.js
```diff
@@ -29,9 +29,9 @@ var latexMathParser = (function() {
   var symbol = regex(/^[^${}\\_^]/).map(function(c) { return VanillaSymbol(c); });
 
   var controlSequence =
-    regex(/^[^\\a-eg-zA-Z]/) // hotfix #164; match MathBlock::write
+    regex(/^[^\\a-eg-zA-Z|]/) // hotfix #164; match MathBlock::write
```
laughinghan: this prohibits parsing of normal pipe characters `|` right?

src/commands/math/advancedSymbols.js
```diff
@@ -322,5 +323,11 @@ LatexCmds.cap = LatexCmds.intersect = LatexCmds.intersection =
 // FIXME: the correct LaTeX would be ^\circ but we can't parse that
 LatexCmds.deg = LatexCmds.degree = bind(VanillaSymbol,'\\degree ','&deg;');
 
+LatexCmds.celsius = bind(VanillaSymbol,'\\celsius ','&#8451;');
+LatexCmds.micro = bind(VanillaSymbol,'\\micro ','&#181;');
+LatexCmds.perthousand = bind(VanillaSymbol,'\\perthousand ','&#8240;');
+LatexCmds.ohm = bind(VanillaSymbol,'\\ohm ','&#8486;');
```
laughinghan: these don't appear to be real LaTeX commands? http://mathb.in/55241

src/commands/math/advancedSymbols.js
```diff
@@ -220,10 +220,11 @@ LatexCmds.opencurlybrace = LatexCmds.lbrace = bind(VanillaSymbol, '\\lbrace ', '
 LatexCmds.closecurlybrace = LatexCmds.rbrace = bind(VanillaSymbol, '\\rbrace ', '}');
 LatexCmds.lbrack = bind(VanillaSymbol, '[');
 LatexCmds.rbrack = bind(VanillaSymbol, ']');
-
+
```
laughinghan: spurious whitespace

src/commands/math/basicSymbols.js
```diff
@@ -414,8 +414,10 @@ var PlusMinus = P(BinaryOperator, function(_) {
 
   _.contactWeld = _.siblingCreated = _.siblingDeleted = function(opts, dir) {
     if (dir === R) return; // ignore if sibling only changed on the right
+    // If the left sibling is a binary operator or a separator (comma, full stop, semicolon, colon)
+    // consider the operator to be unary, otherwise binary
     this.jQ[0].className =
-      (!this[L] || this[L] instanceof BinaryOperator ? '' : 'mq-binary-operator');
+      (!this[L] || this[L] instanceof BinaryOperator || ',.;:'.indexOf(this[L].ctrlSeq) >= 0 ? '' : 'mq-binary-operator');
```
laughinghan: I'm not sure about dot `.`, in math that's usually a decimal point, right? Like `1.0 - 1. - 1`, MathJax at least doesn't consider the minus in that case to be unary: http://mathb.in/55242

laughinghan: also, you could consider using a regex for brevity: https://github.com/mathquill/mathquill/blob/v0.10.1/src/commands/math/commands.js#L415

src/commands/math/basicSymbols.js
```diff
@@ -418,9 +418,10 @@ var PlusMinus = P(BinaryOperator, function(_) {
   _.contactWeld = _.siblingCreated = _.siblingDeleted = function(opts, dir) {
     if (dir === R) return; // ignore if sibling only changed on the right
     // If the left sibling is a binary operator or a separator (comma, full stop, semicolon, colon)
+    // or an open bracket (open parenthesis, open square bracket)
     // consider the operator to be unary, otherwise binary
     this.jQ[0].className =
-      (!this[L] || this[L] instanceof BinaryOperator || ',.;:'.indexOf(this[L].ctrlSeq) >= 0 ? '' : 'mq-binary-operator');
+      (!this[L] || this[L] instanceof BinaryOperator || ',.;:(['.indexOf(this[L].ctrlSeq) >= 0 ? '' : 'mq-binary-operator');
```
laughinghan: interesting. Normally this isn't possible to type, typing `(` or `[` creates a paren group not a single symbol, but they can still be rendered from LaTeX via the API or being pasted in, I guess this is the right thing to do in that situation

laughinghan:
@arnog: wow, you got to it right away, thanks so much! Love the screenshots. Could you add tests (maybe to `css.test.js`) to check for the CSS class or something? (Like https://github.com/mathquill/mathquill/pull/606/commits/37af8e5577ed8d671c3d99ce2b4a2f5d2ebde787#diff-a465877fc47f368b25cea4f30fb694b4 kinda deal?)

src/commands/math/advancedSymbols.js
arnog: Well, it depends which packages you want to support. They're in gensymb (ftp://ftp.dante.de/tex-archive/macros/latex/contrib/was/gensymb.pdf) and MathQuill already supports the \degree macro which is also in this package...

laughinghan: oh, for real? Awesome! Can we get rid of the FIXME comment and group them together and comment about gensym and that link then?

src/services/latex.js
arnog: yes, as a control sequence. However, the standalone | character will be parsed y the VanillaSymbol parser above.

src/commands/math/advancedSymbols.js
arnog: Oh... I see... You thought degree was not standard... Also, not sure what the rest of the comment was about because ^\circ is rendered correctly (although different than the \degree macro, since it's literally a circle displayed as an exponent... Anyway, I'm going to clean this up and put only the changes for the unary operator in this pull request and I'll submit another PR for these other changes.

laughinghan: oh, thanks!

test/unit/css.test.js
```diff
@@ -53,4 +53,29 @@ suite('CSS', function() {
     var testVal = parseFloat(mqF.css('margin-right')) - parseFloat(mqF.css('margin-left'));
     assert.ok(testVal > 0, 'this should be truthy') ;
   });
+
+  test('unary PlusMinus before separator', function () {
+    var mq = MQ.MathField($('<span></span>').appendTo('#mock')[0]);
+    mq.latex('(-1,-1-1)-1,(+1;+1+1)+1,(\\pm{}1,\\pm{}1\\pm{}1)\\pm{}1');
```
laughinghan: these braces don't do anything, `\pm` is a symbol, doesn't take any arguments

test/unit/css.test.js
```diff
@@ -53,4 +53,29 @@ suite('CSS', function() {
...
+    var spans = $(mq.el()).find('.mq-root-block').find('span');
+    assert.ok(spans.length >= 35, 'PlusMinus expression parsed incorrectly');
```
laughinghan: is the nonstrict equality to not be dependent on the specific spans that parens happen to be implemented with? I think using hardcoded indices later makes that moot right? I'm gonna change this to strict equality

test/unit/css.test.js
```diff
@@ -53,4 +53,29 @@ suite('CSS', function() {
...
+    assertUnaryOperator(13, '(-1,-1-1)-1,(+');
+    assertUnaryOperator(16, '(-1,-1-1)-1,(+1;+');
+    assertBinaryOperator(18, '(-1,-1-1)-1,(+1;+1+');
+    assertBinaryOperator(21, '(-1,-1-1)-1,(+1;+1+1)+');
+    assertUnaryOperator(25, '(-1,-1-1)-1,(+1;+1+1)+1,(\pm');
+    assertUnaryOperator(28, '(-1,-1-1)-1,(+1;+1+1)+1,(\pm1,\pm');
+    assertBinaryOperator(30, '(-1,-1-1)-1,(+1;+1+1)+1,(\pm1,\pm1\pm');
+    assertBinaryOperator(33, '(-1,-1-1)-1,(+1;+1+1)+1,(\pm1,\pm1\pm1)\pm');
+  
+  });
```
laughinghan: also you don't clean up by doing `$(mq.el()).remove()`

though really that shouldn't be necessary every goddamn time, especially if you're only adding things to `#mock`

laughinghan: oh, nor did the test above, which is probably what yours is based on

PR #615, opened 2016-04-11
Add support for all amsmath symbols and test/reference.html for validation

arnog:
This test files includes the symbols and commands from the AMS short-math-guide (ftp://ftp.ams.org/pub/tex/doc/amsmath/short-math-guide.pdf) as they are displayed by MathQuill and LaTeX (via codecogs) for reference.


arnog:
Still a work in progress, there are some unfinished sections to cover, but feedback welcome.

arnog:
OK, I've pushed the rest of the changes. The file should now include coverage for all the relevant sections of short-math-guide. For some reason the commit is not showing yet in GitHub, but any day now...

test/reference.html
```diff
@@ -0,0 +1,946 @@
...
+        }
+        
+        t += '</div>';
+        i += 3;
+    };
+    
+    $('#table').html(t);
+    $('.mathquill-static-math').each(function() { 
+        MQ.StaticMath(this);
+        if (this.textContent === '$$\u200B') {
```
laughinghan: this is very dependent on implementation details, I'd recommend using the API, like `MQ(this).latex() === ''`

test/reference.html
```diff
@@ -0,0 +1,946 @@
...
+}
+
+.col5 {
+    -webkit-column-count: 5;
+    -moz-column-count: 5;
+    column-count: 5;
+}
+
+img {
+    opacity: .5;
```
laughinghan: why is this?

laughinghan:
This is awesome! Two moderate concerns:
- The page is very wide, much wider than the Short Math Guide PDF itself even blown up to an equivalent font size, can we get at least the minimum width to be less wide? (If you like wide that's fine, but it should be fluid to smaller widths too within reason.) Width of a PDF at the same font size is usually my rule of thumb
- Every page load of this sends >600 requests to CodeCogs, right? Is that kosher? It's not clear whether or not that conforms to their Usage Policy, which is clearly written with usage on public websites in mind, not local test pages: http://www.codecogs.com/latex/usage.php
    + one other popular alternative is the Google Charts API but that's deprecated apparently? https://github.com/adam-p/markdown-here/issues/261
    + there's also QuickLatex? http://quicklatex.com/
    + I dunno what the right answer here is; other than Google I'd be similarly concerned with spamming requests to any such 3rd-party service
    + other alternatives include rendering locally (MathJax can output SVG, right? Or is there too much that's not supported by MathJax?), or saving all the images (seems unfortunate to bloat our repo, but maybe they're tiny)

laughinghan:
weirder idea: pre-render all the images but instead of keeping them in our repo, upload them to the git asset storage used for storing images in comments and issues image, and use that link

test/reference.html
```diff
@@ -44,48 +58,72 @@
...
 
 .col5 {
     -webkit-column-count: 5;
     -moz-column-count: 5;
     column-count: 5;
-}
-
-img {
-    opacity: .5;
+    display: inline-block;
```
laughinghan: images are replaced elements, aren't they already effectively `inline-block`?

arnog:
I've adjusted the width and the layout. Looking at it printed, it looks pretty good to me, but let me know what you think.

The codecogs TOU say "you can generation up to 3,000* equations a day for testing and non-commercial use". That would seem to cover this local test page, but I get your point.

I've added support for MathJax and commented out the codecogs rendering. The MathJax rendering is pretty good (in fact better than the codecogs one, as it weirdly cuts off the top pixel, at least in SVG), The only notable differences I could spot are no support for \slash, \thickspace, \medspace, \displaylimits, unmatched \begin{?matrix} \end{?matrix} and optional alignment parameters in \cfrac.

test/reference.html
arnog: Ah, yes, much better. Thanks.

laughinghan:
Definitely looking even better, though still getting some width weirdness:
image

test/reference.html
arnog: Yes, I was trying to get the alignment to work right. Fixed it now, don't need this anymore.

test/reference.html
```diff
@@ -907,14 +990,28 @@
...
+       } else {
             t += '<table>';
             while (j < entries.length) {
                 if (entries[j] === sections[i]) {
                     // entry matches the current section
                     t += '<tr>';
                     t += '<td class="mathquill-static-math">' + entries[j + 1] + '</td>';
-                    t += '<td><img src="' + 'http://latex.codecogs.com/svg.latex?' + encodeURI(entries[j + 1]) + '"></td>';
+                    t += '<td class="codecogs"><img src="' + 'http://latex.codecogs.com/svg.latex?' + encodeURI(entries[j + 1]) + '"></td>';
+                    t += '<td class="mathjax">$$' + entries[j + 1] + '$$</td>';
```
laughinghan: does this mean we're relying on the MathJax feature where it searches through every text node on the page for `$$`? Since we're generating the elements it might be a lot faster to call the MathJax API directly on our known elements (like by querying for all elements with the `.mathjax` class and MathJax-ifying them or something)

arnog: I've switched to deferring the typesetting and making it explicit, but I don't see any difference in speed. Not sure why you think it would be faster. Scanning all the nodes to find the $$ is negligible compared to the cost of manipulating the DOM and redoing the layout.

laughinghan: Damn, you're totally right

Han

test/reference.html
arnog: The opacity? I was trying to find some way to make the codecogs rendering less prominent. I've removed the opacity trick and just change the background color now.

arnog:
Added support for the amsmath commands. 

There is one issue remaining with delimiters: they only function with \left + \right commands. Some delimiters were previously defined as non-Bracket (i.e. \lfloor) and were displayed fine standalone. Now, they only display if after a \left. This doesn't affect text entry (which will generate the \left automatically), but it does affect Latex conversion.

Otherwise, things look good. See before and after PDF.

[reference-before.pdf](https://github.com/mathquill/mathquill/files/216164/reference-before.pdf)
[reference.after.pdf](https://github.com/mathquill/mathquill/files/216165/reference.after.pdf)

arnog:
Also worth noting, Symbola is missing a few codepoints: \gvertneqq, \lvertneqq, \ngeqq, \ngeqslant, \nleqq, \nleqslant. Perhaps those could be simulated with a \not + appropriate command?

laughinghan:
Man, this is amazing, a whole mass of new symbols!

There's still some funkiness, but obviously this PR doesn't have to be perfect and fix everything to be merged, it just has to be an improvement. The main thing I'm looking for that would block merging is regressions, and anything else worrisome like the ">600 calls to CodeCogs"-type miscellanea.

- [ ] How can we verify that there are no regressions? I'm thinking we dump the old and new `LatexCmds` to JSON and use a script to compare and make sure the Unicode characters and symbol class (`BinaryOperator` etc) of anything old is the same in the new, or any discrepancies are documented by 9398c4e4baff2e31b4a30528eeb0bdd1d58251ea
- [ ] Can you think of anything else worth double-checking?

----

Stuff I noticed but might not block merging:
- We should comment-out the mention of CodeCogs if we're gonna comment out its usage
- Looks like you made integrals into having `\limits`, so they're displayed like `\sum`, I think that'll conflict with #603
    + Are the new integral/sum/etc signs too big?
- 3.16 Accents&mdash;these are mostly so broken they probably shouldn't be added at all

Stuff I noticed but definitely doesn't block merging:
- MathJax shows up twice for a while? A little Googling suggests maybe it's the PreviewHTML? Can we disable that?
- A lot of the relation symbols like `\approxeq` are `VanillaSymbol`s, they should be `BinaryOperator`s like `=`
- `\vec\imath` and `\hat\jmath` look terrible because of the `text-align: right`

test/reference.html
```diff
@@ -1050,6 +1053,13 @@
             this.innerHTML = '<span style="background:red; color: white; font-weight:bolder; padding-left:.5em; padding-right:.5em;">?</span>';
         }
     });
+    
+    // We've deferred the typeseting on startup. Do it explicitly now on each
+    // elements with a "mathjax" class.
+    $('.mathjax').each(function() { 
+        MathJax.Hub.Queue(["Typeset", MathJax.Hub, this]);
```
laughinghan: if I comment this out, MathJax still auto-renders&mdash;I don't think this commit did anything

src/commands/math/advancedSymbols.js
```diff
@@ -2,43 +2,202 @@
...
+'mid:2223', 'll:226a', 'gg:226b', 'asymp:224d', 'parallel:2225',
+'bowtie:22c8', 'smile:2323', 'sqsubseteq:2291', 'sqsupseteq:2292', 'doteq:2250',
+'frown:2322', 'ni:220b', 'propto:221d', 'vdash:22a2', 'dashv:22a3', 'owns:220b',
+
+// Binary operators
+'mp:2213', 'ominus:2296', 'uplus:228e', 'sqcap:2293', 'ast:2217', 'sqcup:2294',
+'bigcirc:25ef', 'bullet:2219', 'ddagger:2021', 'wr:2240', 'amalg:2a3f',
+
+'cdot:22c5', 'circ:2218', 'div:00f7', 'pm:00b1', /* 'times:00d7', */ 'cap:2229',
+'cup:222a', 'setminus:2216', 'land:2227', 'lor:2228', 'wedge:2227',
```
laughinghan: how are `\land` and `\wedge` distinct? Seems like they're the same class and the same character?

src/commands/math/basicSymbols.js
```diff
@@ -491,3 +422,88 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
...
+}
+
+LatexCmds['#'] = bind(VanillaSymbol, '\\# ', '#');
+LatexCmds['_'] = bind(VanillaSymbol, '\\_ ', '_');
+LatexCmds['|'] = LatexCmds.Vert;
+
+// MathQuill non-standard commands
+LatexCmds.upsih = //W3C/Unicode "upsilon with hook"
+LatexCmds.Upsih = //'cos it makes sense to me
+   bind(Symbol,'\\Upsilon ','<var class="mq-nonSymbola">&upsih;</var>');
```
laughinghan: This probably shouldn't be upright, not `<var>`, right? Just like capital Upsilon (capital Greek letters are customarily upright rather than italicized in math for some reason)

src/commands/math/basicSymbols.js
```diff
@@ -491,3 +422,88 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
...
+    LatexCmds[m[1]] = bind(Variable, '\\' + (m[2] && m[2].length > 0 ? m[2] : m[1]) + ' ', '&#x' + m[3] +';');
+}
+
+for (i = 0; i < BASIC_VANILLA_SYMBOLS.length; i++) {
+    m = BASIC_VANILLA_SYMBOLS[i].match(/([a-zA-Z]+)\/?([a-zA-Z]*):(\w+)/);
+    LatexCmds[m[1]] = bind(VanillaSymbol, '\\' + (m[2] && m[2].length > 0 ? m[2] : m[1]) + ' ', '&#x' + m[3] +';');
+}
+
+LatexCmds['#'] = bind(VanillaSymbol, '\\# ', '#');
+LatexCmds['_'] = bind(VanillaSymbol, '\\_ ', '_');
```
laughinghan: I don't think we can successfully parse `\#` and `\_`...

src/commands/math/basicSymbols.js
```diff
@@ -2,6 +2,35 @@
...
+                              e.g. lowercase greek letters
+                    - Letter
+                        - f
+                - OperatorName
+                - BinaryOperator
+                    - PlusMinus
+                    - Inequality
+                    - Equality
+*/
+
```
laughinghan: 💯

src/commands/math/basicSymbols.js
```diff
@@ -491,3 +423,90 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
...
+for (i = 0; i < VARIABLE_SYMBOLS.length; i++) {
+    m = VARIABLE_SYMBOLS[i].match(/([a-zA-Z]+)\/?([a-zA-Z]*):(\w+)/);
+    LatexCmds[m[1]] = bind(Variable, '\\' + (m[2] && m[2].length > 0 ? m[2] : m[1]) + ' ', '&#x' + m[3] +';');
+}
+
+for (i = 0; i < BASIC_VANILLA_SYMBOLS.length; i++) {
+    m = BASIC_VANILLA_SYMBOLS[i].match(/([a-zA-Z]+)\/?([a-zA-Z]*):(\w+)/);
+    LatexCmds[m[1]] = bind(VanillaSymbol, '\\' + (m[2] && m[2].length > 0 ? m[2] : m[1]) + ' ', '&#x' + m[3] +';');
+}
```
laughinghan: I'm not sure how often the code you're replacing did this, but this would be the perfect spot to support Unicode when pasting in/parsing LaTeX, right? Either by find-replacing all the code points to start with `\u` escape sequences and using the one character, or I guess `String.fromCharCode()` or something

laughinghan:
This really was fantastic work, @arnog!

One more note: I think the ideal solution we'd want to eventually have is to have one file that generates both the `LatexCmds` in source and the documentation/visual test page. Any thoughts on that?

src/commands/math/advancedSymbols.js
arnog: They are indeed synonyms. \land, \lor and \lnot are more commonly used, while \wedge, \vee and \neg are used by lattice theory specialists.

src/commands/math/basicSymbols.js
arnog: Good point. This was the original code: ```bind(Symbol,'\\Upsilon ','<var style="font-family: serif">&upsih;</var>'); ``` but that was probably incorrect as well.

Come to think of it, this probably shouldn't have a control sequence of \Upsilon anymore, since this is now distinct from regular Upsilon. Probably make sense to map it to \upsi, no?

src/commands/math/basicSymbols.js
arnog: ```\#``` seems to work (trying it out in demo.htm) but you're right for \_. I'll comment it out.

arnog:
I like the idea of comparing a JSON dump of LatexCmds (and CharCmds?). I'll look into that.

Integrals (like ```\int```, ```\oint```, etc...) and **big operators** like ```\bigvee``` are of class **SummationNotation**. Should they all be **VanillaSymbols** for now until PR#603 is merged? Seems like they look better for now as a **SummationNotation** until something better comes along with PR#603, no? And are you suggesting we only keep ```\int```, ```\sum```, ```\prod``` and ```\coprod``` as SummationNotation?

Yes, the big operators do look a little big. That should probably be adjusted. Anything you have in PR#603 for this?

I agree about the accents, I'll turn them off for now. It's unfortunate that any expression that uses them will now fail rather than displaying something. It would probably be useful to have a lenient option that would still display as much as possible an expression with Latex it cannot parse fully (like MathJax does).

I've optimized the MathJax rendering to avoid the double render and to generally speed it up.

Good catch on the relation symbols. They should indeed be **BinaryOperator**s.

arnog:
I believe I've addressed all the previous comments. Here is the generated before/after comparison of LatexCmds. Let me know what you think and if there's anything else missing before merging to master.

[collation.pdf](https://github.com/mathquill/mathquill/files/221859/collation.pdf)

src/commands/math/basicSymbols.js
laughinghan: > Come to think of it, this probably shouldn't have a control sequence of \Upsilon anymore, since this is now distinct from regular Upsilon. Probably make sense to map it to \upsi, no?
 
Oh shoot, I was only thinking about backcompat when importing LaTeX, I forgot about interop when exporting. There's no real AMS-LaTeX command for upsilon with a hook, right? The only way for this to interop with MathJax is probably `\unicode{}`, and no way with KaTeX (except Unicode in the source), right? I'm thinking we should just comment this out and note this as a breaking change.

laughinghan:
The `\upsih` thing made me remember, besides LaTeX import backcompat, the other thing to check for is LaTeX output interop. I expect this not to be a big problem, since these are all from the Short Math Guide to LaTeX, right? Still, how can we verify we haven't added anything that isn't real LaTeX or can't be parsed by MathJax or KaTeX? (Or if we have, we should comment them or something.)

Thank you so much for the new comparison! Unfortunately even just the changed section has almost 300 commands (15 or so per page for 19 pages and change), which is just too many to individually look at. I'm looking for outliers, things that might be regressions or interop problems like the `\deg` conflict with `amsmath` or `\ast` vs `\star`, as noted in your commit message in 9398c4e4baff2e31b4a30528eeb0bdd1d58251ea&mdash;that commit message was perfect, actually, summarizing code changes affecting hundreds of commands in like a dozen items. Can we group these by what the change was, like grouping all the symbols who only change was from `VanillaSymbol` to `BinaryOperator` into a section, etc?

Could you also provide the script you're using to generate these, like in a comment or Gist or something? Also if it would make sense, the script you used to generate these commands from the KaTeX `symbols.js`, if that's still where this is all from? How much manual editing did you do afterwards?

I really can't express enough how impressed I am, by the way. Can I get you cookies or something when this is merged?

arnog:
The file ```reference.html``` should be a good indicator regarding the interop of the output since we compare it to MathJax (and codecogs if we enable it). I could add the MathJax rendering for the added commands in the collation document (probably not worth doing for the existing ones, since there are already dozens that are non-standard). Thoughts?

I could try to group the changes in categories, but that will still leave lots of changes to review. I'm not sure there's a way around that. That's how I arrived at the commit message listing the differences.

I didn't use a script to get the symbols from KaTeX, just a RegEx in a text editor. AFAIR I didn't change any of the values afterwards. I'm attaching the source for the collation document to this comment.
[collation.html.txt](https://github.com/mathquill/mathquill/files/222268/collation.html.txt)

src/commands/math/basicSymbols.js
arnog: OK, but ```\upsih``` is not the only command that doesn't interop with AMS-LaTeX or MathJax. For example ```\parallelogram```.

src/commands/math/advancedSymbols.js
```diff
@@ -2,43 +2,202 @@
...
+}
+
+for (i = 0; i < VANILLA_SYMBOLS.length; i++) {
+    m = VANILLA_SYMBOLS[i].match(/([a-zA-Z]+)\/?([a-zA-Z]*):(\w+)/);
+    LatexCmds[m[1]] = bind(VanillaSymbol, '\\' + (m[2] && m[2].length > 0 ? m[2] : m[1]) + ' ', '&#x' + m[3] +';');
+}
+
+LatexCmds.iff = LatexCmds.Longleftrightarrow;
+LatexCmds.implies = LatexCmds.Longrightarrow;
+LatexCmds.enspace = LatexCmds['\\;'];
```
laughinghan: `LatexCmds['\\;']` doesn't appear to exist, nor `LatexCmds[';']`; removing

laughinghan:
> I could add the MathJax rendering for the added commands in the collation document (probably not worth doing for the existing ones, since there are already dozens that are non-standard). Thoughts?

I think I was unclear. I don't care about interop of nonstandard aliases like `\epsiv` or `\xists`, because while MathQuill accepts those when importing LaTeX, MathQuill never exports those; when exporting, those are normalized to e.g. `\varepsilon` and `\exists` respectively, which are valid LaTeX commands that correctly interop with MathJax and KaTeX.

I _am_ worried about e.g. `\parallelogram`, because MathQuill _does_ export that, and it doesn't interop with MathJax and KaTeX, which I guess don't support the `gensymb` package, only `amsmath`. Now, `\parallelogram` in particular wasn't added or worsened by this PR, so we can leave it be, but I want to ensure we don't accidentally add non-interoperable commands, if we do at all I want it to be deliberate. For example, I notice that `\sigmav` used to be an alias for, and normalized to, `\varsigma`, but that's broken by this PR: on this PR, if you type <code>\sigmav </code>, MathQuill will output the non-interoperable `\sigmav`, when it should output an interoperable `\varsigma`. A totally reasonable oversight, but a problem we should fix before merging.

On the other hand, `&upsih;` (`ϒ`, U+03D2) seems to have no exact AMS-LaTeX equivalent, since as you pointed out, `\Upsilon` should in fact be `Υ`, U+03A5. That's a candidate for deliberately outputting a non-interoperable control sequence, though I'm leaning towards leaving it as an alias, since I think reducing our interop gap is more important than maintaining this particular, rather subtle distinction.

> I could try to group the changes in categories, but that will still leave lots of changes to review. I'm not sure there's a way around that. That's how I arrived at the commit message listing the differences.

That commit message is exactly what we need, "lots of changes" being like a dozen groups of changes to review, rather than hundreds in no particular order or organization. Thank you for the source to the collation document, I've modified it so the "Commands changed" section has 9 subsections (and re-generated the dump of the original commands to preserve a little more information): https://gist.github.com/laughinghan/80a399fcd492746d9d29522d180d8fb1

To run it, add `MQ.LatexCommands = LatexCmds;` to `src/publicapi.js` (just like yours required) and [delete `\enspace`](https://github.com/mathquill/mathquill/pull/615#discussion_r59976804) ([patch that does both of these things](https://gist.github.com/laughinghan/e5da7e33194b3900da3c69ec6344d8a6)).

Result: [MathQuill Collation 0bc270f.pdf](https://github.com/mathquill/mathquill/files/222582/MathQuill_Collation_0bc270f.pdf)

It only lists 196 changed commands rather than close to 300 as in yours. I believe this is because while yours treated decimal and hex numeric character entity references as equivalent, yours didn't consider e.g. `&#2229;` and `&cap;` equivalent even though they're parsed as the same Unicode character, which I sidestepped by comparing Unicode characters directly.

By the by, the way I noticed the `\sigmav` problem mentioned above was when I was looking at the "Changes to aliases" subsection:
image

laughinghan:
https://github.com/arnog/mathquill/pull/1 suggests some changes to address things brought up by the new "collation" document, reducing the "Commands removed" section from this:
image

To this:
image

I'm not sure what to do about the remaining removed command, which appears to be because `\vert` used to alias to the normal ASCII pipe character `|`, and you changed it to map to the fancy Unicode character U+2223. KaTeX appears to render both the normal pipe `|` and `\vert` as U+2223, whereas MathJax renders them both as the normal ASCII pipe. They do seem to agree that normal pipe `|` and `\vert` should be aliases of each other.

laughinghan:
Looks like there's some cleanup work left, but the end is in sight and we're set up so that when we're ready to merge, we'll know it with confidence!

arnog:
The previous commit addressed the latest issues raised. After this commit, the collation looks like the attached PDF. Note that \| is indeed defined, but it is an alias to \Vert for consistency with LaTeX (and MathJax and KaTex).

[collation.pdf](https://github.com/mathquill/mathquill/files/224967/collation.pdf)

dominique-unruh:
Currently (i.e., in c41aef4), macros such as \lVert and \rVert do not create brackets. The intended use in TeX is as brackets. (Otherwise, we'd just use \Vert.) I could not find a way to enter them (neither via text entry, not via LaTeX) as brackets.

laughinghan:
Working on the merge conflicts with `master`, jQuery 1.5.2 is causing MathJax to behave funnily where the rendered math is all smooshed together at the end of the section instead of inside the table where it should be (switching back to jQuery 1.7.2 relieves the problem):

screen shot 2017-02-21 at 18 26 02

I tried generating wrapping `$$ ... $$`s but with just one call to `MathJax.Hub.Typeset()` (unlike what we'd tried before which called that many, many times: https://github.com/arnog/mathquill/commit/eb3299bd013d7aaadcd9b02d85b94a7fbb65a4c2#diff-76ea3ebad150f64578dbeb98a5cfcfc1L1291 ), and timed it compared to using jQuery 1.7.2, they both take ~9s, the `<script type="math/tex">` thing doesn't really save us any time, so I'm reverting us to that.

520qzx520:
mathquill How do I add latex symbols that mathquill does not support

PR #617, opened 2016-04-15
Strawman: Custom symbols option via ENV_VAR

laughinghan:
Alternative to #538, allows specifying a `customSymbols` option on a per-field basis.

Because use of the parser now requires the `localLatexCmds` `ENV_VAR` to have a value, we have to make sure no one ever does like `latexMathParser.block.parse()` or anything, so we rename `Parser::parse` to `Parser::_parse` to ensure we're not using it anywhere. This constitutes a large fraction of the changes.

Upon further thought I'm not sure either this or #538 is a good idea at all because it encourages a proliferation of similar-looking dialects of TeX by MathQuill API consumers; instead, perhaps `\embed{}` could be extended to support this use case?

### DIFF
--- a/src/services/latex.js
+++ b/src/services/latex.js
@@ -36,7 +36,7 @@ var latexMathParser = (function() {
       .or(regex(/^\s+/).result(' '))
       .or(any)
     )).then(function(ctrlSeq) {
-      var cmdKlass = LatexCmds[ctrlSeq];
+      var cmdKlass = localLatexCmds.value[ctrlSeq];
 
       if (cmdKlass) {
         return cmdKlass(ctrlSeq).parser();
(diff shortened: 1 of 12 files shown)


src/services/latex.js
```diff
@@ -97,19 +97,73 @@ var latexMathParser = (function() {
...
+ * (More about Concise JSON Schemas: https://git.io/vwI9k )
+ *
+ * This is based on two classifications TeX uses for typesetting:
+ *   - the type of atom ("ord, op, bin, rel, open, close, punct and inner")
+ *       + determines spacing and line-breaking
+ *   - the family of font (normal upright roman, italic, etc)
+ *
+ * https://github.com/Khan/KaTeX/wiki/Examining-TeX#group-types
+ * http://tex.stackexchange.com/a/38984
+ */
```
laughinghan: MathML is, unsurprisingly, unhelpful here. Presentation MathML has separate `<mi>`, `<mn>`, and `<mo>` elements ("identifiers", "numerals", and "operators", resp.), but they're for semantics (which really should be the purview of Content MathML...), not, y'know, "Presentation" like they should be. For font family stuff, all 3 elements have a `mathvariant` attribute; for spacing, `<mo>` has `form`, `lspace`, and `rspace` attributes, which are in theory supposed to have smart defaults, but for example the `form` attribute's smart default doesn't handle unary negation in the case of `1+-2`: http://jsbin.com/cilicu/edit?html,output

laughinghan: Point being: looked at MathML's classifications for atoms; decided they shouldn't influence MathQuill API's classifications.

PR #621, opened 2016-04-19
Merge accessible Mathquill

sclower:
I've been given the go-ahead to submit a PR for merging my fork into master. Please let me know of any blockers.


laughinghan:
Exciting @sclower! Will get back to you

README.md
```diff
@@ -284,6 +288,9 @@ non-italicized when typing the letters without typing a backslash first, like
 3.17 of the Short Math Guide][3.17]) plus some missing trig operators like
 `sech`, `arcsec`, `arsinh`, etc.
 
+You can also specify a speech-friendly representation of the operator name by supplying the operator name, a `|` and its speech alternative (separate multiple words with a `-`). For example, 'sin|sine cos|cosine tan|tangent sinh|hyperbolic-sine'.
```
jwmerrill: What about people that don't specify any autoOperators (so that users need to type a backslash before sin to get the operator form). It seems like MathQuill might want to have a default concept of how the operators that it knows about should be spoken.

src/commands/math/commands.js
```diff
@@ -369,11 +373,11 @@ var SummationNotation = P(MathCommand, function(_, super_) {
   };
 });
 
-LatexCmds['∑'] =
+LatexCmds['?'] =
 LatexCmds.sum =
 LatexCmds.summation = bind(SummationNotation,'\\sum ','&sum;');
 
-LatexCmds['∏'] =
+LatexCmds['?'] =
```
jwmerrill: What happened here, with replacing the capital sigma and capital pi symbols with question marks?

sclower: The symbols are there, but because they’re UTF-8 maybe your viewer is converting them to ANSI?

jwmerrill: Looks like this happened in https://github.com/sclower/mathquill/commit/f53cfb10caacbc1f9bf3b1d31f5589cabb78a199. It seems like a mistake to me.

src/commands/math/commands.js
```diff
@@ -437,7 +455,7 @@ CharCmds['/'] = P(Fraction, function(_, super_) {
 
 var SquareRoot =
 LatexCmds.sqrt =
-LatexCmds['√'] = P(MathCommand, function(_, super_) {
+LatexCmds['v'] = P(MathCommand, function(_, super_) {
```
jwmerrill: Replacing the unicode square root symbol with a v also seems like it might be a mistake. Also happened in https://github.com/sclower/mathquill/commit/f53cfb10caacbc1f9bf3b1d31f5589cabb78a199

README.md
sclower: Agreed. The defaults looked kind of kludged together and I wasn’t keen on breaking something in this first round. @laughinghan your thoughts are welcome.

src/commands/math/commands.js
jwmerrill: I did a second pass scan of the cumulative diff, and it seems like this is the only file and https://github.com/sclower/mathquill/commit/f53cfb10caacbc1f9bf3b1d31f5589cabb78a199 is the only commit where unicode symbols got replaced incorrectly.

jwmerrill: > The symbols are there, but because they’re UTF-8 maybe your viewer is converting them to ANSI?

I wonder if this is an encoding difference between windows and mac or something? This isn't purely a display issue on github, because the symbol is showing up as changed in the diff, and the old symbol displays correctly for me, but the new one does not.

sclower: I will reconvert the file back to UNICODE (I had to change initially because of the bomb causing the minifying to fail) and commit if it works.

jwmerrill: @sclower is it easy for you to find the commit where the byte order mark confusion slipped in? Seems like we'll want to be on the lookout for this issue to avoid more confusion down the road.

sclower: I can, if need be. After that fiasco I've switched to Notepad++ for any 
edits on Windows as you can instruct it to not write the BOM, thus 
making uglifyjs happy.

kschaefe:
So, we have ARIA work that we've done outside of MathQuill and I want to ensure that this feature can be disabled.  I do not want it interfering with what we're already doing in our project.

sclower:
I’m not in favor of providing an option to disable accessibility. What do you feel would conflict?

kschaefe:
We have a custom rendering of math that is already placed in ARIA attributes as appropriate.  Our Math, creates a stack of elements that as such for read-only math:

<span>
<span aria-hidden="true"><mathquill></span
<span style="display: none;">custom math TTS data</span>
</span>

For editable math, we do something a tad more complex, but it essentially follows the same pattern.  With the possibility of many MQ fields on the screen, we need the power to have MQ TTS pieces a stative as possible.

I dislike the idea that MathQuill should solve this with a globally injected aria-block.  Allow the developer to specify what block to use.  Note that we might like to use a different element for each MQ field.  If you have many fields partially completed, maintaining that state per field is important.

Our math parsing algorithm, which produces the TTS text for us, has cases that we think make the math better to understand for our K-12 audience.  The math-to-text algorithm should be replaceable without digging into the guts of MQ.  Allow us to use our own algorithm.  If need be, I could replace that with a no-op algorithm until we could get our custom algorithm compliant with the API for the new MQ aria bits.

I don't want to get into a situation where I can't update MathQuill to get other new features because the ARIA approach taken conflicts with the one that I'm already using.

sclower:
If you maintain an internal MathQuill fork, and if this PR is approved, I recommend the no-op replacement. Specifically, the methods in src/services/aria.js are ultimately responsible for creating and updating the ARIA alert element.

I do see the potential value for adding a flag for disabling ARIA, but only if there is an accessible alternative at hand. As it stands, the company paying me to work on this feature has not asked for a way to disable accessibility, so I am unlikely to write one unless it turns out we need it for some reason, or if it is required for PR approval.

kschaefe:
But you're not building this in an extensible way that make it useful for other developers.  You don't respond to any of the concerns that I raised that offered possible alternative approaches to make it extensible.

Having to whack out huge portions of the MQ base in our fork is not sustainable long-term.

sclower:
That makes sense, and I don’t want to burden you with additional work since you are already developing a MathQuill fork of your own. To keep things simple, we’ll close this PR and keep our accessible fork internal.

As someone who grew up with no accessible math technology, I’m really glad you're working on solving the same kinds of problems we are.. We should compare notes at some point and share thoughts on best practices for accessibility and TTS.

Thanks for the feedback on this approach.

kschaefe:
I don't think you should close this PR based solely on my comments, others may be interested.  We have tackled this road and I was merely raising concerns about having a unilateral MQ approach that would not work in our larger context.  For others, this solution may be perfect.

Unfortunately, our TTS generation is tied to an internal, private bit of math code.  I'd favor working with you to find a solution that enables this PR for the wider audience, but doesn't prevent alternate TTS strategies.

PR #622, opened 2016-04-19

jwmerrill:

[2]: https://web.archive.org/web/20150406091222/https://developers.google.com/mobile/articles/fast_buttons

(Google seems to have lost the original widely-cited "Fast Buttons" article: https://developers.google.com/mobile/articles/fast_buttons )

--

tap highlight box.

### DIFF
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -139,7 +139,7 @@ function getInterface(v) {
       this.__controller.root.postOrder('reflow');
       return this;
     };
-  _.fastClick = function(target, clientX, clientY) {
+  _.clickAt = function(target, clientX, clientY) {
     var ctrlr = this.__controller, root = ctrlr.root;
     var el = document.elementFromPoint(clientX, clientY);
     if (!jQuery.contains(root.jQ[0], el)) el = root.jQ[0];
--- a/src/publicapi.js
+++ b/src/publicapi.js
@@ -139,19 +139,6 @@ function getInterface(v) {
       this.__controller.root.postOrder('reflow');
       return this;
     };
-  _.clickAt = function(clientX, clientY, target) {
-    target = target || document.elementFromPoint(clientX, clientY);
-
-    var ctrlr = this.__controller, root = ctrlr.root;
-    if (!jQuery.contains(root.jQ[0], target)) target = root.jQ[0];
-    ctrlr.seek($(target), clientX + pageXOffset, clientY + pageYOffset);
-    if (ctrlr.blurred) this.focus();
-    return this;
-  };
-  _.ignoreNextMousedown = function(fn) {
-    this.__controller.cursor.options.ignoreNextMousedown = fn;
-    return this;
-  };
   });
   MQ.prototype = AbstractMathQuill.prototype;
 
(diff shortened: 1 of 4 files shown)


jwmerrill:
Supercedes #513

laughinghan:
@jwmerrill Hmmm, looks like the rebase lost the comments on the original commit, including a useful one about the choice of `clientX/Y` over other options like `pageX/Y`. How do you feel about in the future, we fix conflicts by merging `master` into topic branches?

jwmerrill:
Fine by me. Feel free to close this and reopen the other one if you prefer.

jwmerrill:
Also, I'm not 100% sure I resolved the document conflict the way you want it to be resolved. Do you want the docs to talk about "touch taps" or not?

laughinghan:
No it's fine, we'll stick with this one, the document merge was perfect.

laughinghan:
Re @kschaefe <https://github.com/mathquill/mathquill/pull/403#issuecomment-138973223>:
> How does this change interact with other fast-click libraries, such as [jQuery UI Touch Punch](http://touchpunch.furf.com/)?

_TL;DR: they don't interact but you should use this API instead of Touch Punch_

Firstly your Touch Punch wouldn't stand a chance against my HTML5 Shiv, which any good defensive programmer keeps handy...I mean, no, I don't think they really interact, this change just adds API methods, if you don't call them then this PR doesn't change any behavior.

"Touch Punch" in particular looks like it would, narrowly speaking, "just work" with MathQuill without you having to do anything special (like call these API methods), but what it does may not be what you want. It appears to prevent touch swipes from scrolling the container and instead converts touch events one-to-one to mouse events (touchstart/touchmove/touchend to mousedown/mousemove/mouseup). This is exactly what you want for, say, a [jQuery UI Draggable](https://jqueryui.com/draggable/) element, since touch swipes starting from such an element should drag it around and not scroll the page. However, this means that a touch swipe on a MathQuill editable field would, like mouse-dragging, create a mouse selection from touchstart to touchend.

By contrast, on a normal `<input>` or `<textarea>`, touch swipes will scroll the container. Still, maybe "touch swipes" = "mouse selection" is exactly what you want, because selection is important to you and you just have a few math fields dotting a page full of inert text that can be swiped to scroll. However, if you have an expression list filled to the brim with math fields, then swiping a math field would be the only way to scroll, and the only touch gesture that you'd want MathQuill to pay attention to would be a touch tap, which you'd still want to behave like a mouse click.

For that to work, you can't call `.preventDefault()` on touch events [like Touch Punch does](https://github.com/furf/jquery-ui-touch-punch/blob/4bc009145202d9c7483ba85f3a236a8f3470354d/jquery.ui.touch-punch.js#L38) because that's what prevents touch swipes from scrolling the container, instead you have to use the API provided by this PR. (Which is an annoying API&mdash;Touch Punch's `.preventDefault()` call conveniently suppressed the 300ms-delayed legacy mouse event, obviating having to do anything like `.ignoreNextMousedown()`.)

jwmerrill:
New code looks good to me.

kschaefe:
@laughinghan Thanks for the response.  We've got some legacy locked-in items right now and knowing that this will not interact negatively with them is great.  I also appreciate the detail on the Touch Punch evaluation; it will be useful in giving me leverage to update some things.

jwmerrill:
I could have sworn that this PR also added documentation for this method, but I guess it doesn't?

laughinghan:
It never has, haha. Maybe you were just thinking of the commit message? Anyway I'm gonna add docs for this to #623

PR #623, opened 2016-04-19
Refactor documentation from README into Read the Docs

stufflebear:
The new documentation: http://mathquill.readthedocs.org/en/latest/

This breaks apart the content in the README into multiple pages and rewrites a lot of the sections.

This does not yet incorporate any of the documentation from the wiki.


docs/Api_Methods.md
```diff
@@ -0,0 +1,228 @@
...
+
+var firstMQ = MathQuill.getInterface(2);
+firstMQ.MathField(...);
+</script>
+```
+
+This lets different copies of MathQuill each power their own math fields, but using different copies on the same DOM element won't work. `noConflict()` is intended to help you reduce globals.
+
+# Constructors
+
```
kschaefe: It would be good to include the `\\MathQuillMathField` here, since that feature seems to be lost in the documentation.  Not sure there's a better location than the constructor section.

docs/Api_Methods.md
```diff
@@ -0,0 +1,228 @@
...
+Changes the config of the mathField to the new [configuration](http://mathquill.readthedocs.org/en/latest/Config/#setting-configuration).
+
+## dropEmbedded(pageX, pageY, options) **[Experimental](http://mathquill.readthedocs.org/en/latest/Api_Methods/#note-on-experimental-features)**
+
+Insert a custom embedded element at the given coordinates, where `options` is an object like:
+```js
+{
+  htmlString: '<span class="custom-embed"></span>',
+  text: function() { return 'custom_embed'; },
+  latex: function() { return '\customEmbed'; }
```
kschaefe: Should be `\\` for latex return.

docs/Api_Methods.md
laughinghan: good point, though that's not documented anywhere yet I think (except `demo.html` if we're really stretching), so I don't think this blocks merging

stufflebear: Added in https://github.com/mathquill/mathquill/pull/623/commits/727ea47b35acac3a756059ab99eeefcdefd37534

laughinghan:
I'm trying to make <http://docs.mathquill.com/> the canonical URL and I think I did [the right steps](http://read-the-docs.readthedocs.io/en/latest/canonical.html#enabling) in the Admin panel but it doesn't seem to work (no `<link rel="canonical" />` at all on our docs).

laughinghan:
Also, how do we change the sidebar order? Googling seems to only have turned up overwriting the whole `mkdocs.yml` ([StackOverflow], [StackOverflow again]), which doesn't seem right, it seems like there's gotta be an easier way.

Seems like the right order for the sidebar should be something like:
- Home
- Getting Started
- API Reference
- Config Options
- Under the Hood
- Code of Conduct

[StackOverflow]: http://stackoverflow.com/a/30151670/362030
[StackOverflow again]: http://stackoverflow.com/a/27624503/362030

stufflebear:
docs.mathquill.com works for me. I don't think reordering the six options in the sidebar should be a blocker

PR #624, opened 2016-04-20
Add \not parser to handle combinations like \not\ni

eoghanmcilwaine:
Hi @laughinghan, I had a crack at this [before](https://github.com/Learnosity/mathquill/commit/523578166984ee6deded2f41a088627a1ba77ca4) but I've cleaned it up by moving everything into the `\not` parser as you suggested.

cc @karoltarasiuk

### DIFF
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -239,8 +239,42 @@ LatexCmds.bull = LatexCmds.bullet = bind(VanillaSymbol,'\\bullet ','&bull;');
 LatexCmds.setminus = LatexCmds.smallsetminus =
   bind(VanillaSymbol,'\\setminus ','&#8726;');
 
-LatexCmds.not = //bind(Symbol,'\\not ','<span class="not">/</span>');
 LatexCmds['¬'] = LatexCmds.neg = bind(VanillaSymbol,'\\neg ','&not;');
+LatexCmds.not = P(VanillaSymbol, function(_, super_) {
+  // If one of these appears immediately after not, the
+  // parser returns a different symbol.
+  _.suffixes = {
+    '\\ni':       'notni',
+    '\\subset':   'notsubset',
+    '\\subseteq': 'notsubseteq',
+    '\\supset':   'notsupset',
+    '\\supseteq': 'notsupseteq'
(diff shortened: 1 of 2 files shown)


src/commands/math/advancedSymbols.js
```diff
@@ -239,8 +239,42 @@ LatexCmds.bull = LatexCmds.bullet = bind(VanillaSymbol,'\\bullet ','&bull;');
...
+    var succeed = Parser.succeed;
+    var optWhitespace = Parser.optWhitespace;
+
+    // Sort the suffixes, longest first
+    var suffixes = Object.keys(_.suffixes).sort(function(a, b) {
+      return b.length - a.length;
+    });
+
+    // Returns a parser matching any string in array
+    function anyOf(strings) {
```
eoghanmcilwaine: This could be useful as a generic parser method? Or maybe there's a better way to do this that I didn't think of? A big regex might be simpler but it seemed more elegant to do it this way.

src/commands/math/advancedSymbols.js
```diff
@@ -239,8 +239,42 @@ LatexCmds.bull = LatexCmds.bullet = bind(VanillaSymbol,'\\bullet ','&bull;');
...
+  };
+  _.init = function() {
+    return super_.init.call(this, '\\neg ', '&not;');
+  };
+  _.parser = function() {
+    var succeed = Parser.succeed;
+    var optWhitespace = Parser.optWhitespace;
+
+    // Sort the suffixes, longest first
+    var suffixes = Object.keys(_.suffixes).sort(function(a, b) {
```
eoghanmcilwaine: Should I maybe cache all this stuff (including the `anyOf` parser) once during `_.init` rather than doing this work every time `_.parser` is called? I left it here because (a) I didn't see anywhere else where such a specific parser is cached, and (b) I like the idea of being able to change the suffixes dynamically (although this isn't happening at present).

jlchereau:
May I suggest adding the \\in suffix as a complement to https://github.com/mathquill/mathquill/pull/740?

MagicHacker:
Hi! I don't know how parser to work.Can you tell me something about parser? I will be very grateful!

PR #629, opened 2016-05-04
Pin pjs to less >=3.1.0 <5.0.0

jwmerrill:
PJs 5.0.0+ is currently incompatible with MathQuill.

PJs 5.0.0 introduced a new behavior for this.constructor(...) in order
to be compatible with coffeescript classes. The new behavior means
that, if you don't use `new` when you call `this.constructor`, the
object returned from the constructor is an alias of `this` instead of a new object.

This behavior interacts badly with MathQuill's use of PJs to subclass
jQuery, and jQuery's use of `this.constructor` inside `pushStack`:

https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/core.js#L72-L84

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/mathquill/mathquill.git"
   },
   "dependencies": {
-    "pjs": ">=3.1.0"
+    "pjs": ">=3.1.0 <5.0.0"
   },
   "devDependencies": {
     "mocha": "*",


jwmerrill:
Note that if you do 

```
npm install pjs@5.1.1
make clean
make test
```

The unit tests totally bomb. Almost none of them pass, and the test suite doesn't even finish running.

jwmerrill:
The PJs commit that makes the change is this one: https://github.com/jneen/pjs/commit/39b9eaf7b4d4263c1a09b6dd1e9b7a9111feb518

I think it's actually pretty bad behavior to have `this.constructor` quietly mutate `this` if you forget `new`, because it's an easy mistake to make, and you might not notice the consequences very quickly.

jwmerrill:
The commit that changed mathquill's PJs dep is this one: https://github.com/mathquill/mathquill/commit/892d872aefcbcddda51d585eba1ac4c10891f54a

jwmerrill:
To me, this situation also suggests that it's probably a bad idea to use PJs to subclass jQuery.

I don't think subclassing like this is part of jQuery's public API, and subclassing jQuery buys you very little, since you could just do `insDirOf(jQ, dir, el)` instead of `jQ.insDirOf(dir, el)`.

laughinghan:
Thanks for tracking this down! Not sure I agree on whether this is good or bad behavior for Pjs, I'll have to think about it more; the point about subclassing jQuery not seeming worthwhile is well made.

I've repro'd that indeed `npm install pjs@5.0.0 && make clean && make test` breaks the tests, and `4.0.0` instead fixes them, merging.

laughinghan:
I've opened a ticket on Pjs with your minimal test case (thanks for that, too). ES6 subclasses jQuery just fine (http://jsbin.com/xuzuta/edit?html,console,output), I don't think there's any fundamental reason why Pjs additional features should cause it to barf. Whether it's worth fixing is another question.

jwmerrill:
Wait, so according to this link, my minimal test case doesn't actually show
the problem, right?

laughinghan:
> I think it's actually pretty bad behavior to have `this.constructor` quietly mutate `this` if you forget `new`, because it's an easy mistake to make, and you might not notice the consequences very quickly.

You may be right that it's pretty bad behavior, but it's what idiomatic and Coffeescript classes do: http://jsbin.com/lozeqa/edit?html,console,output

laughinghan:
@jwmerrill Huh? That link was with ES6 classes which don't have this problem, suggesting that the problem with Pjs is solveable.

jwmerrill:
I see now, sorry for the noise.

laughinghan:
Actually, everything sucks, the ES6 subclass of jQuery isn't "well-behaved", in that if you try to instantiate the subclass, you instead get a jQuery object that isn't an instance of the subclass (I updated the example). Maybe this isn't a solveable problem.

jwmerrill:
Inheritance: not always as great as it sounds...

PR #633, opened 2016-05-09
playing with dollar

karoltarasiuk:


karoltarasiuk:
In addition to https://github.com/mathquill/mathquill/pull/543 I created this. Work is in progress, but wanted to show you my solution.

It seems to work, only when you have dollars disabled, there's no way to break from \text. I couldn't get ENTER to work as this event doesn't go that deep - any suggestions?

@laughinghan @stufflebear

PR #635, opened 2016-05-11
Fix ReadTheDocs sidebar and code blocks styling

laughinghan:
Code blocks now a reasonable size, and section headers now have little permalink icons when you hover over them, see it here: http://docs.mathquill.com/en/docs.readthedocs/


stufflebear:
Love it. +1

PR #637, opened 2016-05-15
Fix #534 "Better error message if jQuery isn't loaded"

### LINKED ISSUE #534: Better error message if jQuery isn't loaded
> TypeError: _superclass is undefined

https://mathquill.slack.com/archives/mathquill/p1456625051000204

(In my browser I get `Uncaught TypeError: Cannot read property 'prototype' of undefined`, but it's the same deal)

laughinghan:


The-Penultimate-Defenestrator:
I don't see anything that could possibly go wrong because of such a small change.

PR #641, opened 2016-05-31
Fix CircleCI for forks

laughinghan:
People may want to sign up for Circle and Sauce for their public, open
source forks (which I've done for my laughinghan/mathquill fork),
`circle.yml` should be general and work if they do that.

(Instead, it was showing a
```
{"error": "Not found"}
```
in `status.json`.)


mjdesa:
👍

PR #645, opened 2016-06-10
WIP: Diff current screenshots with whats on master.

mjdesa:
TODO:

- [x] Figure out why Edge is weird.
- [x] Log steps in screenshot


circle.yml
```diff
@@ -52,9 +52,10 @@ dependencies:
 
 test:
   override:
-    # Sauce cant connect to Safari unless the server is ran on port 3000 or 4000
+    # Sauce can connect to Safari on ports 3000, 4000, 7000, and 8000. Edge needs port 7000 or 8000.
     # https://david263a.wordpress.com/2015/04/18/fixing-safari-cant-connect-to-localhost-issue-when-using-sauce-labs-connect-tunnel/
-    - PORT=3000 make server:
+    # https://support.saucelabs.com/customer/portal/questions/14368823-requests-to-localhost-on-microsoft-edge-are-failing-over-sauce-connect
```
laughinghan: 👍 👍 👍 for the link

circle.yml
```diff
@@ -137,6 +138,35 @@ test:
...
+        prev=PREV_$file
+        metric_diff=$(compare -metric AE -compose src $prev $file raw_diff.png)
+        composite -alpha on raw_diff.png $prev DIFF_$file
+      done
+
+      for file in $(ls EVERGREEN*); do
+        prev=$(ls PREV_$file*)
+        metric_diff=$(compare -metric AE -compose src $prev $file raw_diff.png)
+        composite -alpha on raw_diff.png $prev DIFF_$file
+      done
```
laughinghan: I think it's preferable to separate the shell commands into separate `circle.yml` lines as much as possible, which is nice because CircleCI times them and displays output from them separately. I see they all share `$dir` which won't get shared across separate `circle.yml` lines, but you can easily add environment variables that all commands will see to `circle.yml`: https://circleci.com/docs/configuration/#environment

script/screenshots.js
```diff
@@ -23,66 +23,94 @@ fs.mkdirSync(allImgsDir);
...
 
     browserDriver.get(url, function(err) {
       if (err) console.log(err);
+      console.log(cfg.browserName,cfg.platform,'get')
       browserDriver.safeExecute('document.documentElement.scrollHeight', function(err,scrollHeight) {
         if (err) console.log(err);
+        console.log(cfg.browserName,cfg.platform,'safeExecute 1')
         browserDriver.safeExecute('document.documentElement.clientHeight', function(err,viewportHeight) {
           if (err) console.log(err);
+          console.log(cfg.browserName,cfg.platform,'safeExecute 2')
```
laughinghan: how about `'get scrollHeight'` and `'get clientHeight'`instead of just `'safeExecute 1/2'`? Also we could probs save a roundtrip by doing one `safeExecute` for `{scrollHeight: document.documentElement.scrollHeight, ...}`, might be worth it if we do a lot of accesses

laughinghan:
@mjdesa Commented a few nitpicks, looks great, ship it whenever you feel like it!

circle.yml
```diff
@@ -137,6 +138,35 @@ test:
       do
         convert $dir/$x/*.png -append $dir/$x.png
       done
+      # Remove all directories in $CIRCLE_ARTIFACTS/img
+      # Currently the pieces aren't kept around. If it's
+      # desirable to keep them around, we should use
+      #    cp -r $dir $CIRCLE_ARTIFACTS/img_pieces
+      # The reason the image pieces aren't currently kept
+      # around is that it was leading to a problem and didn't
+      # want to sink too much time into the problem at the moment.
```
laughinghan: haha, maybe be slightly more specific? IIRC it was that the number of images in our artifacts directory was increasing quadratically, wanna say that?

circle.yml
```diff
@@ -137,6 +138,35 @@ test:
...
+      for file in $(ls PINNED*); do
+        prev=PREV_$file
+        metric_diff=$(compare -metric AE -compose src $prev $file raw_diff.png)
+        composite -alpha on raw_diff.png $prev DIFF_$file
+      done
+
+      for file in $(ls EVERGREEN*); do
+        prev=$(ls PREV_$file*)
+        metric_diff=$(compare -metric AE -compose src $prev $file raw_diff.png)
+        composite -alpha on raw_diff.png $prev DIFF_$file
```
laughinghan: this `raw_diff.png` is temporary, maybe we should clean it up?
screen shot 2016-06-15 at 20 26 34

circle.yml
mjdesa: k

circle.yml
mjdesa: yeah

circle.yml
mjdesa: 👍

laughinghan:
so the next steps I'm imagining would all be towards the goal of posting something useful as a GitHub comment, which I think will make an outsize difference in the utility of this awesome CI screenshot work.

Ideas:
- can we crop the diffs to be just the areas that are different, with a little bit of padding?
- obviously, stabilizing them so that spurious diffs are avoided would be nice
    + is the current measure of difference just the number of pixels different? Any way to also account for how actually different the pixels are (so that e.g. gray and slightly darker gray count for only a little)
- can we reduce the screen width to 6.5 CSS inches (624 pixels) so that more of 'em could fit side-by-side?

mjdesa:
Is there anywhere we can document this so that its persistent outside of this PR?

laughinghan:
How about a master Issue? #649

circle.yml
```diff
@@ -131,13 +132,55 @@ test:
...
+      # including the pieces images. This compounded so that each
+      # iteration of a test run would have all of the images from
+      # the previous test run plus whichever new images were generated.
+      rm -R -- $img_dir/*/
+
+      # Install utility we need
+      npm install -g json
+
+      # Download the latest mathquill artifacts.
+      curl $(curl https://circleci.com/api/v1/project/mathquill/mathquill/latest/artifacts \
```
laughinghan: just noticed the title of this PR mentions the `master` branch, is that what we intended? Looks like we have to specify a branch filter or defaults to "latest build in entire project if omitted", see "Artifacts of the latest Build": https://circleci.com/docs/api/#build-artifacts-latest

mjdesa: Master was intended. Forgot to do that. I'll fix that now. Unless you think it would be preferable to leave things they way they are now.

laughinghan: hmmm, no I think master is right, this current behavior will compare between random PRs, right? We may also eventually want to compare against latest-build-on-this-PR, in addition to against master

PR #646, opened 2016-06-11
Inset diacritics; fix size and position; add many more

laughinghan:
## Inset diacritics

Before, there was no visual indication whether the cursor is "under" a
diacritic (i.e., inside a diacritic block):
Before

After, the diacritic block gets a `box-shadow` that makes it look inset
when the cursor is inside, just like text blocks:
After

## Fix size and position, add many more

| Renderer | Screenshot |
|---|---|
| MathQuill (before) | screen shot 2016-06-09 at 20 58 06 |
| MathQuill (after) | screen shot 2016-06-09 at 21 00 01 |
| MathJax ([MathB.in](http://mathb.in/63306)) | screen shot 2016-06-09 at 21 03 53 |
| KaTeX | screen shot 2016-06-09 at 21 02 36 |
| `pdflatex` (LaTeXiT) | screenshot LaTeXiT |

Compared to other math renderers (including the official (?) `pdflatex`) the
arrow of `\vec` and wavy-waggly symbol of `\tilde` were too big and, for
lowercase letters, too high.

By using the actual Unicode characters for diacritics, they're the right
size and weight (if you try shrinking the arrow so it's the same size as
`\vec` in other renderers, it's too thin; the glyph for the diacritic is
bolded by comparison), and it means diacritics can all use the same CSS
and HTML template.(If we'd continued down the route we were on and used
`.` for `\dot`, it would've been much too low without special-cased
`\dot`-specific CSS.)

With this generic implementation, we add support for all non-stretchy
diacritics mentioned in the Short Math Guide to LaTeX, plus quadruple
dot `\ddddot`. (This actually brings us to parity with the first time I
tried to implement diacritics, on a now dead branch: https://github.com/laughinghan/mathquill/commit/1cadb70a1e053e6efe5415a926882aa5e4ee8d67#diff-239c238d8c090904fa1a9261bf2ef674R112 )

### Replacing `display: block` with `position: absolute`

To vertically stack the diacritical mark above the diacritic block, we
had been using the "two `display: block` elements inside a
`display: inline-block` element" technique that's also used for
fractions. However, when I tried to ensure that the diacritical mark
would take up some space (so as not to overlap with the overline of a
containing square root, for example), the diacritic block became
misaligned from adjacent text in Chrome and Safari (at most but not all
`font-size`s (?!)), see how `A` and `X` don't line up in so many of these:

- Chrome 50 with diacritics implemented using `display: block` and
  `margin`s:
  Chrome 50 with diacritics implemented using `display: block` and `margin`s
- Safari 9.1.1 with diacritics implemented using `display: block` and
  `margin`s:
  Safari 9.1.1 with diacritics implemented using `display: block` and `margin`s

This happens in both Chrome and Safari when using a `margin`, `padding`,
or `height` CSS property on the diacritical mark to take up space.
Using a transparent border actually works around the problem in
Chrome, but alas not in Safari:

- Safari 9.1.1 with diacritics implemented using `display: block` and
  `border-top: solid transparent .Xem`:
  Safari 9.1.1 with diacritics implemented using `display: block` and `border-top: solid transparent .Xem`

To reproduce, checkout this commit and apply the following patch:
https://gist.github.com/laughinghan/811933e6920980df5dc06fb35b3d6b2b

(This has been isolated into a [minimal testcase] and reported to both
[Chromium] and [WebKit].)

[minimal testcase]: http://output.jsbin.com/tivuvov
[Chromium]: https://bugs.chromium.org/p/chromium/issues/detail?id=617763
[WebKit]: https://bugs.webkit.org/show_bug.cgi?id=158441

So that sucks.

Instead I chose to use `position: absolute` on the diacritical mark,
which is actually what I did the first time I tried to implement
diacritics: https://github.com/laughinghan/mathquill/commit/220caffffda374fd89bb7a66ccc6f7b4a5600684

Who knows why, but without any block-level elements inside the diacritic
inline-block, and the right choice of padding (more on this in a moment),
the misalignment is gone in Chrome and much reduced in Safari:

- Safari 9.1.1 with this PR's diacritics implementation:
  Safari 9.1.1 with this PR's diacritics implementation

It's not all unicorns and disruption. With `.25em` instead of `.2em` of padding,
Chrome becomes misaligned again, though like Safari it's much reduced:

- Chrome 50 with this PR's diacritics implementation except with
  `padding-top: .25em`:
  Chrome 50 with this PR's diacritics implementation except with `padding-top: .25em`

So this may be brittle, hence the new testcase in test/visual.html.

Besides less severe misalignment, another advantage of this
`position: absolute` technique over the `display: block` technique is
that we just have to change the one CSS property, `padding-top`, to both
lower the diacritic and reduce the space that it takes up, which we want
to do because...

### Lowercase letters

After shrinking the arrow and tilde, the fact that the diacritical marks
are too high above lowercase letters becomes especially pronounced.
Unfortunately, CSS/JS/DOM offer no good way to find the height in pixels
of a particular character in a particular font, the only techniques on
the Internet that work at all involve drawing the letter to a `<canvas>`
and examining the pixels ([StackOverflow][1], [StackOverflow again][2]),
which, um, no. KaTeX actually [hardcodes a JSON literal in their source][3]
with metrics for each character (?) in each of their fonts. (No idea what
MathJax does.)

[1]: http://stackoverflow.com/q/16816071/362030
[2]: http://stackoverflow.com/q/1134586/362030
[3]: https://github.com/Khan/KaTeX/blob/v0.6.0/src/fontMetricsData.js

You can see that KaTeX, MathJax, and pdflatex (LaTeXiT) do place
diacritics at heights carefully tailored to each character, with
diacritics above uppercase letters not quite the same height as above
tall lowercase letters (`b`, `d`, etc), nor quite the same height as
`i` and `j`:

| Renderer | Screenshot |
|---|---|
| MathQuill |  |
| MathJax ([MathB.in](http://mathb.in/62706)) |  |
| KaTeX |  |
| `pdflatex` (LaTeXiT) |  |

And, diacritics above capital Roman letters are slightly higher than
above capital Greek letters:

| Renderer | Screenshot |
|---|---|
| MathQuill |  |
| MathJax ([MathB.in](http://mathb.in/62707)) |  |
| KaTeX |  |
|`pdflatex` (LaTeXiT) |  |

Overall, though, the height of diacritics above uppercase letters and
above tall lowercase letters is basically the same, and the height above
short lowercase letters is basically all the same; the only other height
significantly different from those two typical heights is above `t`,
which is somewhere in the middle, annoyingly.

So, I chose to simply use a hashset of all short lowercase Roman and
Greek letters (created at load time from hardcoded lists), and have a
special case for `t`, and assume anything else is tall. This does the
right thing for uppercase Roman and Greek letters, and is the right
height for `1`,`2`, `&`, `%` etc, but it doesn't take into account that
non-letter symbols aren't slanted, so the diacritic looks too far to
the right, and it's also the wrong height for comma `,` and other short,
non-letter symbols.

Anyway, way better-looking and more faithful to real LaTeX than before
for letters, which is what matters.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -489,20 +489,54 @@ LatexCmds.nthroot = P(SquareRoot, function(_, super_) {
   };
 });
 
+// the Short Math Guide to LaTeX calls these "Accents"
+//   (Section 3.17: http://tinyurl.com/gq4c4od )
+// but Wikipedia says only some diacritical marks like acute and grave are accents
 var DiacriticAbove = P(MathCommand, function(_, super_) {
   _.init = function(ctrlSeq, symbol, textTemplate) {
     var htmlTemplate =
-      '<span class="mq-non-leaf">'
-      +   '<span class="mq-diacritic-above">'+symbol+'</span>'
-      +   '<span class="mq-diacritic-stem mq-inset">&0</span>'
+        '<span class="mq-diacritic mq-non-leaf">'
+      +   '<span class="mq-diacritic-above">'+symbol+'&nbsp;</span>'
+      +   '<span class="mq-non-leaf mq-inset">&0</span>'
       + '</span>'
     ;
 
     super_.init.call(this, ctrlSeq, htmlTemplate, textTemplate);
   };
+  _.reflow = function() {
+    var block = this.ends[L], allLow = !block.isEmpty(), anyTs = false;
+    if (allLow) block.eachChild(function(cmd) {
+      if (!(cmd.ctrlSeq in LOW_LETTERS)) allLow = false;
+      if (cmd.ctrlSeq === 't') anyTs = true;
+    });
+    this.jQ.toggleClass('mq-diacritic-low', allLow)
+           .toggleClass('mq-diacritic-t', anyTs);
+  };
+  var LOW_ROMAN_LETTERS = 'acegmnopqrstuvwxyz';
+  var LOW_GREEK_LETTERS = 'alpha gamma eta iota kappa mu nu rho sigma tau chi omega varphi epsilon varepsilon pi varpi varsigma upsilon digamma varkappa varrho psi'.split(' ');
+
+  var LOW_LETTERS = {};
+  for (var i = 0; i < LOW_ROMAN_LETTERS.length; i += 1) {
+    LOW_LETTERS[LOW_ROMAN_LETTERS.charAt(i)] = 1;
+  }
+  for (var i = 0; i < LOW_GREEK_LETTERS.length; i += 1) {
+    LOW_LETTERS['\\'+LOW_GREEK_LETTERS[i]+' '] = 1;
+  }
 });
-LatexCmds.vec = bind(DiacriticAbove, '\\vec', '&rarr;', ['vec(', ')']);
-LatexCmds.tilde = bind(DiacriticAbove, '\\tilde', '~', ['tilde(', ')']);
+// Unicode "Combining Diacritical Marks"
+LatexCmds.grave = bind(DiacriticAbove, '\\grave', '&#x0300;', ['grave(', ')']);
+LatexCmds.acute = bind(DiacriticAbove, '\\acute', '&#x0301;', ['acute(', ')']);
+LatexCmds.hat = bind(DiacriticAbove, '\\hat', '&#x0302;', ['hat(', ')']); // aka circumflex
+LatexCmds.tilde = bind(DiacriticAbove, '\\tilde', '&#x0303;', ['tilde(', ')']);
+LatexCmds.bar = bind(DiacriticAbove, '\\bar', '&#x0304;', ['bar(', ')']); // aka macron/overline
+LatexCmds.breve = bind(DiacriticAbove, '\\breve', '&#x0306;', ['breve(', ')']);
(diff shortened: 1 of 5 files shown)


mjdesa:
So I was playing around trying to get a feel for things and came across something possibly peculiar. 

Typing `\tilde+SPACE` yields
unspecified

If I then type `x` the tilde drops down to the left, inside of the box.
unspecified-1
 
Is this intended?

laughinghan:
@mjdesa: yes that's deliberate, see the "Lowercase letters" subsection of the PR description

laughinghan:
@mjdesa: yes that's deliberate, see the "Lowercase letters" subsection of the PR description:

screen shot 2016-06-17 at 14 31 45

mjdesa:
Just read the section. I could be missing something, but in particular, I think it's odd that the box-shadow passes through the tilde. Less that it moving down and to the left.

laughinghan:
Hmmm, you think it looks too broken? We could try lowering the line-height:
http://jsbin.com/jazefi/edit?html,css,output

Han

mjdesa:
Whatever it was in here

AkshayaRathinavel:
I have merged your PR into the latest mathquill. It adds hat symbol. But it is not rendering properly.
Please refer the screenshot added.
screenshot-staging open-sunbird org-2018 10 05-16-37-44

Please provide suggestion to render the symbols correctly

PR #648, opened 2016-06-14
Add double prime symbol via `"` or \dprime

eoghanmcilwaine:
Includes @stufflebear 's suggestion to call it "dprime" to be consistent with unicode-math.

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -247,6 +247,7 @@ LatexCmds.f = P(Letter, function(_, super_) {
 LatexCmds[' '] = LatexCmds.space = bind(VanillaSymbol, '\\ ', '&nbsp;');
 
 LatexCmds["'"] = LatexCmds.prime = bind(VanillaSymbol, "'", '&prime;');
+LatexCmds.dprime = bind(VanillaSymbol, '"', '&Prime;');
 
 LatexCmds.backslash = bind(VanillaSymbol,'\\backslash ','\\');
 if (!CharCmds['\\']) CharCmds['\\'] = LatexCmds.backslash;


laughinghan:
Since `"` is not synonymous with `\dprime` in real LaTeX ( http://mathb.in/63742 ), I think we should only use `\dprime` here and not have `"` at all, or if you want typing double-quote `"` to result in this, do `CharCmds['"']` has the only mention of `"`. (Also, can you remind me why to normal primes `''` doesn't suffice?)

laughinghan:
We could even do something like `CharCmds['"'] = bind(LatexFragment, "''");` so that typing `"` would insert two normal primes

eoghanmcilwaine:
I've removed the " character binding and updated the pull request. Just checking that it's ok to include `\dprime` which seems to be specific to the unicode-math latex package?

The ability to render this as a single symbol is useful to us because (a) the cursor traverses it in one movement and (b) the appearance (size and spacing) is a bit better than either a plain " or two 's.

laughinghan:
> Just checking that it's ok to include `\dprime` which seems to be specific to the unicode-math latex package?

Augh...no, I don't think this is okay. This is really frustrating. Compat with MathJax seems to be mutually exclusive with compat with "real LaTeX" here, to the extent that such a thing exists.

To display `&Prime;` (aka `\u2033`) in MathJax we have 2 options, either directly include the Unicode character in the LaTeX source, or use MathJax's [custom `\unicode{}` extension][unicode] (`\unicode{x2033}`), which nobody else supports.

The best claim to the throne of "real LaTeX" is pdfTeX ([the default TeX engine in most LaTeX distributions][default]), which doesn't support rendering [arbitrary Unicode characters by hex code][arbitrary] but in this case, [by using the `utf8x` package][utf8x], I was able to render `&Prime;` in [LaTeXiT] (based on pdfTeX) by directly including the Unicode character in the LaTeX source. The newer TeX engines XeTeX and LuaTeX have full UTF-8 support and should also support directly including Unicode in the source. (More about the various TeX engines: [overview], [detailed].)

The `unicode-math` package doesn't actually add Unicode support to anything, it only works on XeTeX and LuaTeX anyway, it's actually just a convenience package providing backslash commands for every Unicode math symbol.

So now what?

Well, directly including the Unicode character in the LaTeX source has broad support, including even "real LaTeX" (pdfTeX) if `utf8x` is used. Let's go with that, and let's not do `\dprime`, which doesn't work in MathJax and would gratuitously add a dependency on `unicode-math` for XeTeX and LuaTeX.

This isn't ideal, of course. For one thing, if we're assuming support for directly including Unicode in the source, why not just do that for all symbols? I guess the answer is that where possible we use the commands supported by pdfTeX out-of-the-box, and only fallback to assuming `utf8x` if absolutely necessary (gotta document this). For another, apparently `utf8x` is [discouraged] in preference to `utf8` because `utf8x` relies on the unmaintained `ucs` package, but I tried `utf8` and it didn't support `&Prime;`, so we're stuck with the deprecated dependency.

**TL;DR** I'll submit a PR to use the Unicode character instead of `\dprime`.

[unicode]: http://docs.mathjax.org/en/latest/tex.html#unicode-support
[default]: https://en.wikipedia.org/wiki/PdfTeX
[arbitrary]: http://tex.stackexchange.com/questions/289982/how-to-type-unicode-in-plain-tex/290011#comment703254_290032
[utf8x]: https://en.wikibooks.org/wiki/LaTeX/Special_Characters#Extending_the_support
[LaTeXiT]: https://www.chachatelier.fr/latexit/
[overview]: https://www.sharelatex.com/blog/2012/12/01/the-tex-family-tree-latex-pdftex-xelatex-luatex-context.html
[detailed]: http://tex.stackexchange.com/q/36
[discouraged]: http://tex.stackexchange.com/a/13070

eoghanmcilwaine:
Thanks @laughinghan, we can definitely work with this. Appreciate the detailed information too!

PR #651, opened 2016-06-21
Symlink from docs/{index,README}.md to README.md in package root, not the other way around

laughinghan:
...because `README.md` in the package root being a symlink was causing problems for `npm pack` (npm/npm#3310).

Ugh, of course duplicate direct symlinks are the last thing I try, just when I've given up. I think the reason that ff628549fc985ee6ec5827cd202e77b0723319d1 didn't work was that it was a *double* symlink, from `docs/index.md` to `docs/README.md` which in turn was a symlink to `../README.md`. By contrast, on this PR, both `docs/index.md` and `docs/README.md` are direct symlinks to `../README.md`, the README in the package root.

Live at http://docs.mathquill.com/en/docs.readthedocs/


stufflebear:
LGTM

Issue #35, opened 2011-05-02
jQuery 1.4.3 required


JoeProgram:
Hi Guys!

I'm working at an educational company, and we really love what you're doing with MathQuill.  I've been digging around in the codebase for a few days, and I noticed that around line 102 in rootelements.js, you use:

   }).bind('selectstart.mathquill', false ).blur();

This notation is not available until jQuery 1.4.3, though the MathQuill homepage says you can use jQuery 1.4+  

Joe

PR #652, opened 2016-06-22
Fix jQuery dependency to be 1.5.2+

laughinghan:
Use `.bind()`/`.unbind()` instead of `.on()`/`.off()`, which were added in
jQuery 1.7:
https://blog.jquery.com/2011/11/03/jquery-1-7-released/

And in our tests, use `jQuery.extend()` instead of the second argument to
`jQuery.Event()` where you can pass in an object and its properties are
set on the newly created `Event` object, which was added in jQuery 1.6:
http://api.jquery.com/category/events/event-object/

And in our docs, bump our minimum required jQuery version from 1.4.3+
to 1.5.2+, because we need some kind of focus/blur-related jQuery bugfix
added in 1.5.2 `[0]`. I've added a test case that fails with jQuery 1.5.1
but passes with jQuery 1.5.2 and onwards (hopefully; tested with 1.7.2,
which was what our tests had been using, and 1.12.4, the latest 1.x).

```
[0]: Best guess: https://bugs.jquery.com/ticket/7340
     Other possibilities:
       https://blog.jquery.com/2011/03/31/jquery-152-released/
```

Finally, downgrade the version of jQuery we test with from 1.7.2 to
1.5.2, so we won't accidentally use new features like `.on()` again.

I guess we've come a long way since #35 when we first documented the
1.4.3+ requirement.

### DIFF
--- a/docs/Getting_Started.md
+++ b/docs/Getting_Started.md
@@ -2,7 +2,7 @@
 
 Download [the latest release](https://github.com/mathquill/mathquill/releases/latest) or [build from source](Contributing.md#building-and-testing).
 
-MathQuill depends on [jQuery 1.4.3+](http://jquery.com), we recommend the [Google CDN-hosted copy](http://code.google.com/apis/libraries/devguide.html#jquery).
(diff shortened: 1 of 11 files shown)


laughinghan:
(a cherry-pick of) this is included on #653

PR #653, opened 2016-06-23
New 2-step release process: script/{prep,push}-release.sh

### LINKED ISSUE #563: Include ultra-bare-bones quickstart.html in the release tarball

laughinghan:
Steps:
1. On a clean tree on `master`, add a new entry to `CHANGELOG.md` in the
   same format as the other entries, then run `script/prep-release.sh`
   to do everything that can be done locally: bump version, build,
   package as tarball + zipfile, commit, tag.
2. After double-checking the build/package/commit/tag, run
   `script/push-release.sh` (with a GitHub access token env variable)
   to automatically publish to NPM, push to GitHub, create a new GitHub
   Release, and cleanup the tarballs etc.

To make that happen, the commits on this PR:
- Inject the version string into the JS and CSS header comments during build
- Simplify `.gitignore` and `Makefile` to not deal with tarballs anymore
- Add a `"files"` array to `package.json`, to configure `npm publish`/`npm pack`
- Fix usage of jQuery to support versions as old as 1.5.2 (obviates #652)
- Add a `quickstart.html` to make it easy to sanity-test the release package
  before publishing (this uses jQuery 1.5.2; closes #563)
- Reformat `CHANGELOG.md` and add `script/{prep,push}-release.sh` to
  automatically figure out from the changelog the appropriate version bump,
  commit message, tag annotation, and GitHub Release description.

See individual commit messages for more details.

Resolves #436: this PR adds publishing to the NPM registry to the release process.


Makefile
```diff
@@ -15,6 +15,14 @@ ifneq ($(shell node -e 'console.log("I am Node.js")'), I am Node.js)
...
 endif
 
+# stupid BSD vs GNU https://developer.apple.com/library/mac/documentation/OpenSource/Conceptual/ShellScripting/PortingScriptstoMacOSX/PortingScriptstoMacOSX.html#//apple_ref/doc/uid/TP40004268-TP40003517-SW20
+ifeq (x, $(shell echo xy | sed -E 's/(x)y/\1/' 2>/dev/null))
+	# BSD
+	SED_IN_PLACE = sed -i ''
+else
+	# GNU
+	SED_IN_PLACE = sed -i
+endif
```
laughinghan: This doesn't work. Apparently the version of `sed` running on CircleCI (GNU sed 4.2.2, reportedly) supports the `-E` flag just like BSD sed, but does NOT support the `-i ''` syntax for in-place editing without backup files like BSD sed. (Which isn't documented in its manpage, nor any GNU sed manpage I could find online, only [this blogpost](http://blog.dmitryleskov.com/small-hacks/mysterious-gnu-sed-option-e/).)

This amazingly perfectly illustrates the subtle but important difference between feature detection and feature inference. [Apple's docs] aren't wrong, their example only uses `-E` so checking `-E` and falling back to `-r` (for old versions of GNU sed, I guess) works perfectly. However, I attempted to infer from the `-E` check behavior of `-i`, which evidently doesn't work.

[Apple's docs]: https://developer.apple.com/library/mac/documentation/OpenSource/Conceptual/ShellScripting/PortingScriptstoMacOSX/PortingScriptstoMacOSX.html#//apple_ref/doc/uid/TP40004268-TP40003517-SW20

I could learn my lesson and directly check whether `-i` works...but that requires several lines and temporary files and shit, so instead, I'm just gonna use Jeanine's original feature inference which worked fine (and still works, I just tested on my laptop and while SSH-ed into CircleCI): it assumes that support for `-r` means GNU's `-i` syntax, and lack of support for `-r` means BSD's `-i ''` syntax (82d50c8f6a6aa77942ab5a932a61b36ae70332f2). Yes, this would break if BSD adopted GNU's `-r` flag without changing its `-i ''` syntax, but `-E` matches `grep -E` and stands for "extended regexes" whereas `-r` doesn't really make any sense, so it seems less likely for BSD to add support for `-r` than for GNU to support `-E`.

(The version of `sed` on CircleCI:
```
$ sed --version
sed (GNU sed) 4.2.2
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
and Paolo Bonzini.
GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Be sure to include the word ``sed'' somewhere in the ``Subject:'' field.
```
)

jwmerrill:
Approved, but there are merge conflicts.

PR #654, opened 2016-06-23
Add `maxDepth` option to limit nesting level

eoghanmcilwaine:
Applies to latex added via keyboard input, pasting, public methods
and initialization. When this option is set, math cannot be added more
than `maxDepth` MathBlocks deep in the math tree.

### DIFF
--- a/src/services/latex.js
+++ b/src/services/latex.js
@@ -77,6 +77,10 @@ Controller.open(function(_, super_) {
   _.exportLatex = function() {
     return this.root.latex().replace(/(\\[a-z]+) (?![a-z])/ig,'$1');
   };
+
+  optionProcessors.maxDepth = function(depth) {
+    return (typeof depth === 'number') ? depth : undefined;
+  };
   _.writeLatex = function(latex) {
     var cursor = this.notify('edit').cursor;
 
@@ -85,7 +89,7 @@ Controller.open(function(_, super_) {
 
     var block = latexMathParser.skip(eof).or(all.result(false)).parse(latex);
 
-    if (block && !block.isEmpty()) {
+    if (block && !block.isEmpty() && block.prepareInsertionAt(cursor)) {
       block.children().adopt(cursor.parent, cursor[L], cursor[R]);
       var jQ = block.jQize();
       jQ.insertBefore(cursor.jQ);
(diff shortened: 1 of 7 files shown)


src/services/latex.js
```diff
@@ -77,6 +77,10 @@ Controller.open(function(_, super_) {
   _.exportLatex = function() {
     return this.root.latex().replace(/(\\[a-z]+) (?![a-z])/ig,'$1');
   };
+
+  optionProcessors.maxDepth = function(depth) {
+    return (typeof depth === 'number') ? depth : undefined;
```
kschaefe: If the default in options for maxDepth was `Number.MAX_SAFE_INTEGER || Number.MAX_VALUE` then the undefined checks used elsewhere are unnecessary.  I mean we could choose another arbitrarily large number, but those are built into JavaScript.

eoghanmcilwaine: Hi @kschaefe, the undefined checks are really there for performance reasons. If maxDepth defaults to some large number, then we'll be doing an unnecessary breadth-first search on every insertion.

eoghanmcilwaine:
Thanks for the review @kschaefe. I've just updated my PR to rebase off master.

Yeah, your example of a fraction inside parentheses inside a square root would be three levels deep. (The contents of the fraction would be at level 4.) `MathBlock` is used as the level boundary, so the minimum level is 1 because the whole mathquill editor is enclosed in a `RootMathBlock`. `isStyleBlock: true` nodes are ignored as they aren't MathBlocks.

When this option is set, you won't be able to add any math deeper than the specified nesting level. If you try to type, paste, or initialize deeper math, the portion that is deeper than the allowed limit will be truncated. Same for entering math programatically via public methods, or wrapping existing content in a new block (a newly created fraction / parenteses etc.); if any portion of the wrapped math is now too deep, it will be removed.

We added this to our fork to stop users deliberately causing JS errors by entering crazily deep content which can't be handled by the parser. (It fails with `Uncaught RangeError: Maximum call stack size exceeded`.) Initializing mathquill with deep math fails at a lower nesting level than you can achieve by typing the math in keypress by keypress, so it's useful to enforce this option if you are saving math and showing it again later.

src/services/latex.js
kschaefe: Gotcha.  Yes, a fast-exit from the function is preferable to a search every time.  It totally worth that complexity then.

kschaefe:
Thanks for the response on how this works and confirming how style blocks are handled.  Seems like a good change.  We sometimes have students enter gobbledygook and this would help prevent that.

PR #656, opened 2016-06-24
[Mathspace] Reflow when closing brackets (Fixes #655)

### LINKED ISSUE #655: `Edit` handler is not called when typing a mismatched bracket
**To reproduce:**

* open up http://mathquill.com/
* type: (3,4] into the field labelled *Type math here:*.
* Latex below the field looks like:
image

MathQuill's internal state has been updated correctly at this point (to render `\left(3,4\right]`), but the edit handler wasn't called.

ajhyndman:
Trigger a reflow event when closing a bracketed expression

If this doesn't happen, an edit event won't be triggered when we write an
unmatched expression like `(3, 4]`.  This means that our handler won't
know about the mismatched bracket until we type something else.

Fixes #655

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -558,8 +558,8 @@ var Bracket = P(P(MathCommand, DelimsMixin), function(_, super_) {
         Fragment(cursor[side], cursor.parent.ends[side], -side) // me and ghost outside
           .disown().withDirAdopt(-side, brack.parent, brack, brack[side])
           .jQ.insDirOf(side, brack.jQ);
-        brack.bubble('reflow');
       }
+      brack.bubble('reflow');
     }
     else {
       brack = this, side = brack.side;
--- a/test/unit/publicapi.test.js
+++ b/test/unit/publicapi.test.js
@@ -304,6 +304,23 @@ suite('Public API', function() {
       });
     }
   });
+  
+  suite('edit handler', function() {
+    test('fires when closing a bracket expression', function() {
+      var count = 0;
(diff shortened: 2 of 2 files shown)


laughinghan:
Nice! But, what about the existing cases where `reflow` _is_ triggered, will this result in unnecessary duplicate calls? I notice that `.closeOpposing()` (to which you added a call to `.bubble('reflow')`) is called in `_.createLeftOf()` just before a conditional call to `.bubble('reflow')`, and when deleting a side, which doesn't need to `.bubble('reflow')` because deleting always bubbles reflow anyway in `src/services/keystroke.js`. What about simply moving the `.bubble('reflow')` call in `_.createLeftOf()` out of the conditional, so it just unconditionally does `.bubble('reflow')` when closing a bracket expression by typing?

laughinghan:
From Slack:
> Are there any unit tests around event behaviour?
> I found a small bug, and am wondering the best way to demonstrate a proposed fix.

Yep, they're [in `test/publicapi.js`](https://github.com/mathquill/mathquill/blob/master/test/unit/publicapi.test.js#L168-L306), though this should be much simpler than those, something like:
```js
var count = 0;
var mq = MQ.MathField($('<span>').appendTo('#mock')[0], {
  handlers: {
    edit: function() {
      count += 1;
    }
  }
});
mq.typedText('(3, 4');
var countBeforeClosingBracket = count;
mq.typedText(']');
assert.equal(count, countBeforeClosingBracket + 1);
```

ajhyndman:
I added a unit test, and tweaked the fix.  Thanks for the tip, I extracted the bubble call from the conditional statement.  That looks cleaner to me..

Issue #508, opened 2016-02-17
We export but can't import \left\langle...\right\rangle


laughinghan:

dominique-unruh:
Also `\left|...\right\rangle` and `\left\langle...\right|` should be supported (in case a fix of the above would not automatically fix this).

These combinations are very important for typesetting formulas in quantum mechanics / quantum computing (kets and bras), without support for this idiom, basically no quantum mechanics can be typeset.

PR #659, opened 2016-06-30
Support for brackets \langle,\rangle, \lVert,\rVert

dominique-unruh:
Support for brackets \langle,\rangle (latex parsing was broken, #508).
Added brackets \lVert,\rVert (renders as || ||, used, e.g., for norms).

I also added a simple test case for each of them.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -684,14 +684,14 @@ LatexCmds.left = P(MathCommand, function(_) {
     var succeed = Parser.succeed;
     var optWhitespace = Parser.optWhitespace;
 
-    return optWhitespace.then(regex(/^(?:[([|]|\\\{|\\langle|\\lVert)/))
+    return optWhitespace.then(regex(/^(?:[([|]|\\\{|\\langle\b|\\lVert\b)/))
       .then(function(ctrlSeq) {
         var open = (ctrlSeq.charAt(0) === '\\' ? ctrlSeq.slice(1) : ctrlSeq);
 	if (ctrlSeq=="\\langle") { open = '&lang;'; ctrlSeq = ctrlSeq + ' '; }
 	if (ctrlSeq=="\\lVert") { open = '&#8741;'; ctrlSeq = ctrlSeq + ' '; }
         return latexMathParser.then(function (block) {
           return string('\\right').skip(optWhitespace)
-            .then(regex(/^(?:[\])|]|\\\}|\\rangle|\\rVert)/)).map(function(end) {
+            .then(regex(/^(?:[\])|]|\\\}|\\rangle\b|\\rVert\b)/)).map(function(end) {
               var close = (end.charAt(0) === '\\' ? end.slice(1) : end);
 	      if (end=="\\rangle") { close = '&rang;'; end = end + ' '; }
 	      if (end=="\\rVert") { close = '&#8741;'; end = end + ' '; }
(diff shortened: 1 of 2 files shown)


laughinghan:
Wow, fantastic, this wasn't so bad. Let's add `\b` to the regexes so as not to match `\left\langlerfish x \right\ranglerfish` (and add test cases for such), and then I think this is ready to merge!

laughinghan:
Out of curiosity, do you happen to have any idea why `\lVert` and `\rVert` even exist? Either seems to work on either side of `\left` and `\right`. Kerning, maybe?

dominique-unruh:
The \b is added now.

dominique-unruh:
> Out of curiosity, do you happen to have any idea why \lVert and \rVert even exist? 

The difference is relevant if used without `\left` / `\right`.
In TeX's math mode, tokens (and blocks) can have different categories (such as "open", "close", "binary relation", etc).
`\lVert` and `\rVert` are the `\Vert` symbol, with the category "open" and "close" respectively.
This affects the spacing when they are used without `\left` or `\right` (i.e., in their non-scaling form).
Instead, one can also write `\mathopen\Vert` and `\mathclose\Vert`. 
However, when using `\left` and `\right`, TeX knows that they are opening and closing parentheses, so in that case it does not matter whether one uses `\left\rVert` or `\left\Vert`.
(Same situation with, e.g., `|`. You shouldn't write `|x|` for absolute value, but instead `\mathopen|x\mathclose|` or the more readable `\lvert x\rvert`.)

Fortunately, all this is not relevant for MathQuill, since MathQuill always uses `\left` and `\right`. But when one wants more finetuned formulas (`\left...\right` often gives parentheses that are too big), then these differences become crucial.

laughinghan:
Fascinating. When are `\left` and `\right` too big? I think ideally MathQuill would always render and output the perfect size, unless that's just too hard to determine algorithmically.

dominique-unruh:
The thing is that both MathQuill and Latex can't do much more than take the bounding box and pick the size of the enclosing parentheses a little bigger.

In my opinion, this ends up often being too big.
For example:

I like to make them a bit smaller (using `\Bigl` `\Bigr` in this case):

(Especially in big formulas, the parentheses otherwise grow too fast.)

Or here it's the other way around:

(using `\left` `\right`)
And:

(using `\bigl`)
The latter makes it easier to see the nesting.

Usually, just using `\left` and `\right` gives quite adequate results (especially compared to lower quality formula typesetting in various office suites), but if one wants to make the formulas look really good, then there seems to be no way around some tweaking of sizes and spacings.

Finally, it can also be that for technical reasons you cannot use `\left` and `\right`. This is the case when you have a multiline expression, and the `\left` and `\right` would end up on different lines. Latex cannot handle this. So one needs to pick the size manually.

kikoramathquill:
Mathquill doesn't still import langle. I use function latex() to get latex. Then import this latex, mathquill doesn't show anything.
latex from mathquill : \left\langle1235\right\rangle

It I remove \b from 'optWhitespace.then(regex(/^(?:[([|]|\\\{|\\langle\b|\\lVert\b)/))', it can import.

dominique-unruh:
Indeed. `\left\langle1235\right\rangle` is not parsed by MathQuill, but `\left\langle 1235\right\rangle` is. The reason is that `\b` matches the end of a word only, and according to Javascript's regular expression semantics, `langle12345` is one word. (While for LaTeX, 12345 cannot be part of a word.)

But we cannot just remove `\b` because it was introduced to avoid confusion with, e.g., `\langlerfish` (see above).

To fix this, we need to replace every `\b` by `(?![a-zA-Z])` (this is a negative lookahead that only matches when no letter follows). Then things seem to work.

dominique-unruh:
I created a fix, see  #787.

kikoramathquill:
Thanks. It works well.

PR #662, opened 2016-07-05
Support for user-defined macros

dominique-unruh:
Adding the possibility for used-defined macros.

Rationale: some domain specific math slows typing down a lot. (For example, in quantum mechanics, a "ket" would have to be typed as "|\Psi\rangle", instead of something short like "\bra\Psi".)

MQ.addMacro("example","x^2") will define a macro \example that gets replaced by x^2 automatically (only in typed input, not in parsed latex)

Put \cursor somewhere in the macro code to indicate where the cursor should be after insertion.
E.g., MQ.addMacro("abs","\left|\cursor\right|")

### DIFF
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -345,16 +345,26 @@ var CursorPlaceholder = P(Symbol, function(_, super_) {
     cursor should be after insertion.  E.g.,
     bind(Macro,"abs","\\left|\\cursor\\right|")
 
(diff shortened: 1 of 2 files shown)


dominique-unruh:
This similar to #538, with the following major differences:
* With this one, exported latex contains only standard TeX symbols (macros are for input only), with #538 the exported latex becomes potentially incompatible to other MathQuill or latex implementations.
* With #538 one can add arbitrary new symbols, with this one one can only implement macros that can be expressed in terms of already existing latex symbols.

PR #667, opened 2016-07-18
Fixed string variable declaration

emily-c:
There was a semicolon instead of a comma adding the `string` variable to window instead of to the local scope.


laughinghan:
Your profile pic hurts my eyes. It's a pretty cool effect, actually, apart from being horrifying

PR #672, opened 2016-08-20
Fix the output of latex() for \class commands.

kschaefe:
Used the \textcolor command as a template for correcting the \class
output.  Added tests to latex for \class to prove correctness and for
\textcolor because it was missing.


kschaefe:
Looking at the CI output, I don't understand what failed.  There seems to be an issue with the PNG diffs, but I didn't touch that code.  Any ideas?

PR #673, opened 2016-08-23
Fix binary op in style block

kschaefe:


kschaefe:
`8\\textcolor{red}{-}4-1` results in:
screen shot 2016-08-23 at 1 25 39 pm

The red minus should have the same spacing as the black minus.  This change fixes the bug.

laughinghan:
Brilliant

PR #676, opened 2016-09-16
upgrade \vee, \wedge to Binary Operator

italoc-84:


laughinghan:
Checks out, MathJax does:

screen shot 2017-02-16 at 20 30 42

http://mathb.in/127577

PR #679, opened 2016-09-30
Add \overarc and \overleftrightarrow

kschaefe:
A small addition to the command list for \overarc and \overleftrightarrow.

Note that \overleftrightarrow has some possible issues with font size changes.  These are the same as the \overleftarrow and the \overrightarrow.  See full details in the commit.


laughinghan:
Sweet! Can you add a line to the visual test page? CI will automatically test in all the browsers for us :)

kschaefe:
No problem.  I'll add some visual checks and update.

kschaefe:
So the current arrows are not in the visual test page.  I manually filled in a textfield there with the `\overleftarrow` and `\overrightarrow`.  Is there a spot where you'd like this in the visual test page.

laughinghan:
Adding to the end of LaTeX examples should be fine. Thanks, really appreciate this!

ryan-codingintrigue:
@laughinghan @kschaefe What needs done to get this PR merged? I can help out with fixing conflicts/adding visual tests if required?

kschaefe:
Yes, the visual tests are the outstanding piece.  I'd welcome the help.

Given that we're already using these in our fork, my work time has been too jammed up to get to the visual checks for this merge to go forward.

There was no visual checks for the existing \overline, \overleftarrow, and \overrightarrow.  The new tests should include those, as the "correct" cases for visual comparison.  \overarc and \overleftrightarrow are the new pieces and should behave similarly to the original set.

ryan-codingintrigue:
Hey @kschaefe thanks for taking a look! I've [filed a PR](https://github.com/CarnegieLearningWeb/mathquill/pull/24) on your branch which should bring this up-to-date with master to resolve the conflicted files - plus added the visual tests for both the new commands and the existing line commands

kschaefe:
@ryan-codingintrigue You need to sign the CLA.

ryan-codingintrigue:
Thanks @kschaefe, done.

kschaefe:
@laughinghan Since you're merging pull requests today.  I think that is now ready.

laughinghan:
Doesn't appear to actually work in IE8:

screen shot 2017-07-25 at 10 19 53 am

src/css/math.less
```diff
@@ -372,5 +386,34 @@
...
+          top: -0.34em;
+        }
+      }
+      &:before{
+        -moz-transform: scaleX(-1);
+        -o-transform: scaleX(-1);
+        -webkit-transform: scaleX(-1);
+        transform: scaleX(-1);
+        filter: FlipH;
+        -ms-filter: "FlipH";
```
laughinghan: is this syntax right? the docs I found look nothing like this: https://msdn.microsoft.com/en-us/library/ms532992(v=vs.85).aspx

src/css/math.less
```diff
@@ -372,5 +386,34 @@
...
+      }
+      &:after {
+        display: block;
+        position: relative;
+        top: -2.3em;
```
laughinghan: This only works if there's text inside, not if there's anything bigger, like a fraction inside, right?

src/css/math.less
```diff
@@ -372,5 +386,34 @@
       filter: FlipH;
       -ms-filter: "FlipH";
     }
+    &.mq-arrow-both {
+      vertical-align: text-bottom;
```
laughinghan: would be nice to have some tests of this next to fractions and stuff (like `\frac{1}{2} + \overarrow{abc}`) to make sure this aligns correctly when the line box is tall

src/css/math.less
kschaefe: Probably, but that's the same problem that `\overleftarrow` and `\overrightarrow` have.  They are very susceptible to font size and line height changes.

We should probably have a reflow handler that updates those offsets programmatically.

src/css/math.less
kschaefe: Good call.

src/css/math.less
kschaefe: No idea.  I copied the syntax for `\overleftrightarrow` from `\overleftarrow` and `\overrightarrow`.  I didn't check IE8...to be honest that never would have occurred to me.

PR #681, opened 2016-10-03
Add option to use multiplication symbol instead of cdot

ctlusto:
- Also adds the option to use the division symbol instead of `/`


ctlusto:
Would somebody mind rerunning this test or helping me interpret what's going on here. Looks like it timed out and I'm not sure why. Thanks!

laughinghan:
Dang it's unfortunate that you have to hardcode that into `MathBlock::chToCmd()` because `LatexCmds` can't depend on options.

I'm going to rename these to `typingSlashWritesDivisionSymbol` and `typingAsteriskWritesTimesSymbol` to be more specific about when these symbols are being used (also, it's called `\times` not `\multiplication`), cool?

PR #684, opened 2016-10-18
Fix typo (closes #683)

controversial:

PR #685, opened 2016-10-18
fix typo in api_methods

cedrickhoury:
Removed # from ID


controversial:
I just did this in #684, so I'll close this for now.

PR #687, opened 2016-10-18
Fix copying math in Safari

### LINKED ISSUE #686: Copying from mq input is broken in Safari 10
Repro:

1. Go to http://mathquill.com in Safari 10
2. Type an expression into the input box
image
3. Press cmd-c to copy
4. Your paste buffer is empty.

jwmerrill:
Fixes #686. I have no idea why `box-sizing: border-box` would cause
problems here; I discovered this by bisecting the problem to

058297a18e17e4aba09f738


laughinghan:
Nice work!

PR #690, opened 2016-11-02
Line-through command for selected text 

sabmath:
I want to have a command to add line-through on the selected text. I did it by DOM elements(span) but it is not good idea because when i am using Undo Redo function (working with Latex code stored in an array) i lose line-through from text./


laughinghan:
@sabmath There doesn't appear to be any code or commits here relating to line-through?

sabmath:
@laughinghan actually i am finding selected DOM elements (vars ,spans , ..) and then adding line-through style to it, and if it has line-through i remove that style...
But i wanna do it with mathquill any command to do it ... Like what i did in photo on "y" .
mmmm

sabmath:
I Latex , we can use \sout command for it.

PR #692, opened 2016-11-12
`substituteKeyboardEvents` config option

laughinghan:
Lets API consumers override the keyboard events abstraction with a
wrapper around `saneKeyboardEvents()` that can intercept events and ignore
them or do stuff like insert `ans` before a `+` is typed.


laughinghan:
Merging because this has already been reviewed at https://github.com/desmosinc/mathquill/pull/30

Issue #65, opened 2012-03-01
Mathquill not functionnal in android browsers


Narsil:
Mathquill is not usable in (android at least) mobile browsers:

Tested on mathquill.github.com.

Phone: Android Nexus S (4.0)

Browser: Default one
Can type stuff, but cannot backspace, rendering is ok.

Browser: Chrome
Cannot  type anything which is problematic.

This is probably linked to all the keyboard event stuff going on with mathquill.

nscozzaro:
I have a Pixel 2 phone with Chrome and the Mathquill demo on Mathquill.com doesn't work for me. I can click on the box and the browser automatically zooms to the box, but then when I type, the box doesn't respond at all. Suggestions?

ajhyndman:
@nscozzaro, @otizzz Are you able to reproduce this bug with the latest version of MathQuill off of master?

I did some research yesterday, and I believe this issue was fixed in: https://github.com/mathquill/mathquill/pull/694

nscozzaro:
@ajhyndman is there a link to a site using the latest version that I can try?

bmintz:
@nscozzaro how about http://mathquill.com

nscozzaro:
@bmintz as I said above, I tried it on Mathquill.com. 
I made a video of the issue: https://youtu.be/HanO60UzZF8
You'll notice that delete works, but nothing else. At the very end it unexpectedly started to work after accidentally pressing the return key (the blue arrow). But then I refresh the page and it doesn't work against

ajhyndman:
Yeah, I'm not sure why but http://mathquill.com doesn't see to be running the latest version.

This should have the latest code:

http://many-worlds.glitch.me/mathquill/mathquill/branch/master/test/demo.html

nscozzaro:
Yes that's working for me, thanks! (besides the fact that the square root looks rather wonky). 
I'll update to the latest version of Mathquill and hopefully that solves my issue.

ajhyndman:
Awesome!

I'm going to close this issue.  If anyone runs into it again, please feel free to re-open or create a new issue.

mgaouar:
Hi,

I am using Mathquill latest 0.10.1 and I have the same issue on Android devices, I can reproduce on mathquill.com as well, using the latest Android 8 device.

Any update on this?

Thanks,

ajhyndman:
@mgaouar Are you able to provide some sample code that reproduces this issue?

[...]

Issue #643, opened 2016-06-07
Chrome mobile - strange behavior of input box


kyvaith:
Basing on information from issue #65, I should expect MathQuill to not to work on Chrome browser on mobile device, and indeed - after tapping MathQuill input box, device native virtual keyboard shows up, but non of keys pressed with it shows up in input box. What I discovered is after attaching physical keyboard to Android device with USB OTG, and typing at least one letter or digit with it and disconnecting it, virtual keyboard starts to work as excepted. Is it known behavior, or is it something new?

laughinghan:
News to me. Are you willing to help debug this? Can you listen for keydown,
keypress, keyup, and input events on MathQuill's textarea and record the
sequence of them that's fired when typing with the virtual keyboard vs the
physical keyboard? Also, can you record the sequence of them fired on a
normal, non-MathQuill textarea?

Depending on how deep you're willing to dive in on this, this stuff is
solely the responsibility of saneKeyboardEvents.util.js, which is heavily
commented and hopefully helpful.

Han

scaljeri:
I'm experiencing the same behaviour, Chrome on Android shows weird behaviour if you try to type using the native keyboard. Although it doesn't seem to work, you can type whatever you want, but it doesn't show, but if you hit `enter` it suddenly works. 

This behaviour can be reproduced on the mathquill demo site. Is this a known issue yet ? 

I'm also very interested in a temporarily solution ? 

Cheers

ajhyndman:
I've been playing with this a little today.

My impression is: until the `'Enter'` key is pressed, the `typedText` checker is not running.  This is the only way MathQuill is getting text out of the textarea, since key events all return 227 keycodes in Chrome on Android, so only after the first `'Enter'` keypress does the MathQuill input start working.

![image](https://user-images.githubusercontent.com/11449340/33817546-34fbb9c8-de94-11e7-9404-e9ec6df90128.png)

ajhyndman:
It does appear that the only unique property of `'Enter'` is that it fires a `'keypress'` event.

I tested locally, and #694 does fix this issue for me.

As a side-note, that PR doesn't appear to be included in the version of MathQuill running on MathQuill's homepage?  Maybe it's time for an update?

ajhyndman:
I'm closing this issue (and #65) as fixed by #694.

If anyone runs into this again, please feel free to reopen or create a new issue!

Issue #293, opened 2014-04-22
On Mac, {Cmd,Alt} + arrow key makes us treat selection as typed text


laughinghan:
To repro: on Mac, select stuff, like an `x`, then press `Cmd-Right`. If nothing unexpected happens, type something, like a `y`.

Expected: the cursors moves left, or nothing happens, and then the `y` is typed normally.

Observed: in Firefox, after pressing `Cmd-Right`, the selected `x` would be overwritten with `$x$`. In Webkit, after typing `y`, `$x$y` would be typed after the cursor.

On Mac, `{Cmd,Alt}-{Up,Down,Left,Right}` and `Ctrl-{A,E,F,B,N,P}` are all cursor movements, and hence clear the selection without deleting the text and then put the cursor on one side of the text. In Firefox, all keys fire the `keypress` event (see [UnixPapa](http://unixpapa.com/js/key.html), [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=968056)), which cause `saneKeyboardEvents.util.js` to check the textarea for unselected text and assume it was just typed. In Webkit, the unselected text is also there, we just don't see it until the next time something is typed.

This problematic because if a textarea has `x` selected, and then a user types an `x` vs. pressing `Cmd-Right`, there's [no robust way](https://bugzilla.mozilla.org/show_bug.cgi?id=968056#c6) to tell them apart.

Issue #135, opened 2012-12-18
In Safari, typing over a selection fails


laughinghan:
On mathquill.com, at least (which we just updated to the new `master`, right?).

Steps-to-repro: type something, hit Shift-Left a few times, type something else.

Expected behavior: whatever you selected is overwritten by whatever else you typed.

Observed behavior: Nothing. Happens. \o/

@jayferd noooes our bulletproof `textarea.js` not so bulletproof after all

(Reported by Eli in Safari 5.1.5 (6534.55.3), repro'd by Han in Safari 5.1.7 (6534.57.2).)

PR #694, opened 2016-11-16
Handle missing keypress event

ollanta:
I was butting heads with #65 and related issues, and found [evidence](http://stackoverflow.com/questions/21055270/javascript-keypress-event-not-raised-on-android-browser) that the keypress event isn't sent on some mobile browsers. I tested listening for it, and true enough, it wasn't.

Here I've added a listener for the `keyup` event in case `keypress` is missing. I'm not sure if this is a good solution to the problem, but it fixed it in my case.

### DIFF
--- a/src/services/saneKeyboardEvents.util.js
+++ b/src/services/saneKeyboardEvents.util.js
@@ -177,6 +177,10 @@ var saneKeyboardEvents = (function() {
 
       checkTextareaFor(typedText);
     }
+    function onKeyup(e) {
+      // Handle case of no keypress event being sent
+      if (!!keydown && !keypress) checkTextareaFor(typedText);
+    }
     function typedText() {
       // If there is a selection, the contents of the textarea couldn't
       // possibly have just been typed in.
@@ -235,6 +239,7 @@ var saneKeyboardEvents = (function() {
     target.bind({
       keydown: onKeydown,
       keypress: onKeypress,
+      keyup: onKeyup,
       focusout: onBlur,
       cut: function() { checkTextareaOnce(function() { handlers.cut(); }); },
       copy: function() { checkTextareaOnce(function() { handlers.copy(); }); },
--- a/test/unit/saneKeyboardEvents.test.js
+++ b/test/unit/saneKeyboardEvents.test.js
@@ -33,6 +33,25 @@ suite('saneKeyboardEvents', function() {
     el.val('a');
   });
 
+  test('normal keys without keypress', function(done) {
+    var counter = 0;
+    saneKeyboardEvents(el, {
+      keystroke: noop,
+      typedText: function(text) {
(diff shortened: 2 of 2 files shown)


laughinghan:
Yes, I guess one of our assumptions in `saneKeyboardEvents` is that text input will always be preceded by a `keypress`; this change is a noop if that assumption were true, and if it isn't I can see how it would fix it, and I can't see how this could cause any harm.

laughinghan:
@ollanta: fascinating find. For posterity, what device, OS, and OS version did you use to reproduce?

Would you mind checking a few more things?
- So `keypress` isn't fired&mdash;is `input` still fired, on your device at least? What about `beforeinput` or passé `textInput`?
- Is the `.value` of the textarea actually not updated until after `keyup` finishes? In other words, is the deferral `checkTextareaFor(typedText)` necessary, or would synchronously calling `typedText()` work?
- Does any of this vary by the key being pressed---e.g. typing letters works correctly, but Space and Enter don't firing `keypress`?

laughinghan:
And I apologize for taking 1.5 months to respond&mdash;I will absolutely respond within a few days next time, I promise! And always feel free to poke me [on our Slack](http://slackin.mathquill.com)

ollanta:
@laughinghan No problem! We ran a modified build in the meantime, but it's nice to have the change approved :)

Regarding your questions:
- I could reproduce it on a Galaxy S6 running Android 6.0.1 and Chrome 55.0.2883.91, so fairly new. But we did see it on other devices as well.
- `input` is still fired on my device, I'm getting `keydown - textInput - input - keyup`. No `beforeinput`.
- The value is updated when the `keyup` event is sent! So just calling `typedText` seems to work :) I think I didn't really understand the comment for `checkTextareaFor`, and thought it safer to use it.
- Letters, numbers and Space all work the same (that is, they don't work since they don't send a `keypress`), but Enter does send a `keypress` event! So you could work around the issue by starting all input with Enter, and later input will work as expected (mirroring this comment https://github.com/mathquill/mathquill/issues/643#issuecomment-245569985)

laughinghan:
Thanks for responding! Fascinating findings.

The Enter thing in particular is interesting, we're not gonna be able to take advantage of that somehow I don't think, but that definitely helps diagnosing similar issues in the future (since bug reporters may be trying different keys from maintainers trying to repro).

The `input` thing I think we _can_ take advantage of. Back in 2014, I considered relying more heavily on `input` (currently `saneKeyboardEvents` revolves around `keydown` and `keypress`), but hesitated because of how `textInput` seemed promising but was then killed. `input` seems pretty reliable today, though, and could potentially handle #135 and #293 while being less hacky than #294: https://github.com/mathquill/mathquill/issues/293#issuecomment-42062807

I'll try to work on that soon!

PR #700, opened 2016-11-26
Modify Symbol to fix text() output for functions

### LINKED ISSUE #699: LaTeX parsing of parens is inconsistent
If you have Mathquill parse "sin(x)" and then have it convert that to LaTeX, you get "\sin(x)". When the user types in "sin(x)" and have Mathquill convert that to LaTeX, you get "\sin\left(x\right)". The latter seems more correct, and the .text() function certainly behaves better with the latter.

bengolds:
In autoUnitalicize, we keep track of which SymbolCommands are functions (sin, log, etc) and make sure that they produce proper .text() output. There are lots of pathological cases, but this is a good temporary fix. 

This works for \left( and \right), but fails for normal parentheses. If we decide to not fix #699 , then this should be modified to work for normal parentheses.

Closes #537.

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -29,7 +29,7 @@ var Variable = P(Symbol, function(_, super_) {
         text = text.slice(1, text.length);
       }
       else if (text[text.length-1] == ' ') {
-        text = text.slice (0, text.length-1);
+        text = text.slice (0, -1);
       }
     } else {
       if (this[L] && !(this[L] instanceof Variable)
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -24,13 +24,22 @@ var Variable = P(Symbol, function(_, super_) {
   };
   _.text = function() {
     var text = this.ctrlSeq;
-    if (this[L] && !(this[L] instanceof Variable)
-        && !(this[L] instanceof BinaryOperator)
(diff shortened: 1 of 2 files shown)


laughinghan:
Amazing! Mind adding some tests? I think the tests for.text() should probably go here https://github.com/mathquill/mathquill/blob/master/test/unit/autoOperatorNames.test.js

drlippman:
Just out of curiosity, I see this closed my PR but I'm not 100% sure why. Is it because it conflicts? Or does this somehow enable the behavior I was hoping for?

laughinghan:
@drlippman: what PR of yours was closed by yours? Don't see any links to any other PRs here?

Man it's not good that there's many possible PRs you could mean I gotta deal with this.

drlippman:
The original comment for this PR said "closes #537"

bengolds:
I just checked -- I uh, don't know what I was thinking when I put that it closed it. It's relevant, but I don't think it closes it at all. Lemme try and fix that.

drlippman:
No worries. Glad I asked :) In your comments, you said "In my mind, sin x should get output as sin(x) anyway", which is essentially what #537 was about. Is there any easy way for me to patch that in to this revised code?

bengolds:
This change is only about the .text function, which is for outputting the typed expression in a format that can be evaluated in math.js. I don't know if you're using the .text() function, but if you are, it should be pretty mergeable with what you built, if I understand your PR correctly.

PR #706, opened 2016-12-12
Demote font files to not be executable

laughinghan:
chmod -x src/font/*

I'm pretty confused why they were all marked executable in the first
place


laughinghan:
For posterity, private Slack convo:
> @ajhyndman:
> Just curious, was there an issue this was resolving?
> It looks like it’s not a regression, maybe git persisted the permissions from earlier versions of the files?
> @laughinghan:
> no, but things that aren't meant to be executed should probs not be marked executable
> yeah I think so

PR #707, opened 2016-12-20
Fix some broken unit tests.

bengolds:
In typing and autoOperatorNames tests, the global MQ.config was being overwritten instead of the local mq.config. This caused breakage down the line in other tests that needed to use operators.

### DIFF
--- a/test/unit/autoOperatorNames.test.js
+++ b/test/unit/autoOperatorNames.test.js
@@ -76,7 +76,7 @@ suite('autoOperatorNames', function() {
 
   suite('override autoOperatorNames', function() {
     test('basic', function() {
-      MQ.config({ autoOperatorNames: 'sin lol' });
+      mq.config({ autoOperatorNames: 'sin lol' });
       mq.typedText('arcsintrololol');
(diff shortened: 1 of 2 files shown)


bengolds:
Do I need to worry about the CircleCI thing? Says I don't have a username there.

laughinghan:
Thanks, both of you! Don't worry about CircleCI, I didn't configure it correctly for forks (by default they suppress env vars for builds of forks because you could submit a PR that adds `echo SAUCE_ACCESS_KEY` to `circle.yml`), should be fixed now.

This is a really good find, I think this is why unit tests on the latest master commit have been failing [on my "Many Worlds" prototype](https://many-worlds.gomix.me/mathquill/mathquill/commit/3468fe962d48a38879cef366e133661b0de65f0a/test/unit.html) even though they've been passing locally: Gomix runs Linux whereas I'm on Mac so the `cat test/unit/*.test.js` concatenates in the unit tests in a different order, and the global side effects of `MQ.config()` was breaking the tests depending on the order they're run in. (Indeed, [unit tests on this PR are passing](http://many-worlds.gomix.me/mathquill/mathquill/pull/707/test/unit.html).)

PR #710, opened 2017-01-05
Reflect the custom classes back to the latex to preserve original input

pstjvn:
Using custom classes works excellently, but the `latex()` command executed as a getter returns the latex with the class'es escape and the content in the curly braces missing: 

field.latex('\\class{my-class-name}{\\fraq{}{}}'); // visualized correctly.
field.latex(); // returns 'class\fraq{}{}'

This commit fixes the reflection and the returned result from the getter is the same as the input.


kschaefe:
This is a duplicate of #672.

pstjvn:
Oh, I see now..

kschaefe:
Your fix is smaller than mine, but I have added unit tests as well.  Do my unit tests work with your fix?

pstjvn:
Yes, the test is passed successfully.

Issue #689, opened 2016-10-28
Publish with module system


kantianethics:
Would be great to be able to use in browserify without shimming.

ashutoshrishi:
+1

isubastiCadmus:
+1

PR #714, opened 2017-01-17
Added AMD module compatibility

Niels-NTG:
This is a simple fix that makes MathQuill usable as AMD dependency.

Without this MathQuill appears to be an empty Object when you want to use it in something like webpack.


Niels-NTG:
This will solve issue #689

CrazyClicker:
Still on hold?

PR #723, opened 2017-02-16
Ci.cleanup

laughinghan:
Major changes:
- throughout circle.yml, use the `json` cli tool
- fix the Sauce Connect retry-ing to only retry on failure rather than always
- on Sauce, use human-readable build name similar to on Circle; in particular, mention the CircleCI build number
- when applicable keep around pieces of screenshots in `imgs/pieces/`, and file away the baseline imgs (formerly the `PREV_*` imgs) in `imgs/baseline/`
- screenshot filename mentions version of browser used (which changes over time for evergreen browsers)
- `script/screenshots.js` now uses the "promise chain"-style of `wd`, much fewer callbacks

Also lots of minor improvements to comments, screenshots script error logging, test names on Sauce, etc.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -39,69 +39,74 @@ dependencies:
   cache_directories:
     - ~/sauce-connect
   pre:
-    - "test $SAUCE_USERNAME && test $SAUCE_ACCESS_KEY
-       # Sauce Labs credentials required. Sign up here: https://saucelabs.com/opensauce/"
     - ? |-
-        {
-          mkdir -p ~/sauce-connect
-          cd ~/sauce-connect
-          if [ -x sc-*-linux/bin/sc ]; then
-            echo Using cached sc-*-linux/bin/sc
-          else
-            time wget https://saucelabs.com/downloads/sc-latest-linux.tar.gz
-            time tar -xzf sc-latest-linux.tar.gz
-          fi
-          # Sauce Connect randomly fails so try twice https://git.io/vPN8v
-          time sc-*-linux/bin/sc --user $SAUCE_USERNAME --api-key $SAUCE_ACCESS_KEY \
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -31,171 +31,204 @@
 # this file is based on https://github.com/circleci/sauce-connect/blob/a65e41c91e02550ce56c75740a422bebc4acbf6f/circle.yml
 # via https://circleci.com/docs/browser-testing-with-sauce-labs/
 
+machine:
+  environment:
+    json: node_modules/.bin/json
```
desa: Whats this do?

circle.yml
```diff
@@ -31,171 +31,204 @@
...
+        test -e ~/sauce_was_ready && exit
+        time sc-*-linux/bin/sc --user $SAUCE_USERNAME --api-key $SAUCE_ACCESS_KEY \
+          --readyfile ~/sauce_is_ready
+        test -e ~/sauce_was_ready && exit
+        echo 'ERROR: Exited twice without creating readyfile' \
+          | tee /dev/stderr > ~/sauce_is_ready
+        exit 1
       :
         background: true
+    - test -x $json || npm install json
```
desa: we might consider using `jq` instead since I think that comes with the standard Ubuntu install. Maybe not though

desa:
I have to say reading your commits is always enjoyable.

circle.yml
laughinghan: `jq` is fucking dope!

Fixed by https://github.com/mathquill/mathquill/pull/723/commits/50abeb4a2e2486b8f7bd8a6a5ebbc3125b8dea1c

laughinghan: how do you know all the cool JSON utilities?!

PR #726, opened 2017-02-25
CI: Add Many-Worlds build link to Commit Status

laughinghan:
So [Many-Worlds] is this thing I hacked together on Gomix that can fetch,
build, and serve any branch, PR, or commit of MathQuill, on demand. For
example, if you go to:

  http://many-worlds.gomix.me/mathquill/mathquill/branch/master/test/unit.html

then it will fetch the latest `master` branch, run `make test`, then
delegate to a static file server to serve `test/unit.html` and
`build/mathquill.test.js` and all that good stuff.

[Many-Worlds]: https://gomix.com/#!/project/many-worlds

Currently, though, manually formatting that URL is the only way to use
it. What would be nice is if an auto-generated URL were generated per
branch or PR or commit or something and available on GitHub, so you
could just click a link to try out a PR. We could automatically post a
comment on new PRs, but that would spam PR authors with notifications.

A better, quieter way is to use the GitHub Commit Status API:
https://developer.github.com/v3/repos/statuses/

This lets us quietly attach a link to every commit built by CircleCI
without spamming anyone with notifications, and lets you try out
intermediate commits on a PR; not just the tip, every push.

Authentication
--------------

CircleCI doesn't provide us with the token they use for their status
update, probably for good security reasons, but this is exactly what the
MathQuillBot GitHub bot user is for.

`$GITHUB_STATUS_API_KEY` is a [personal access token] for MathQuillBot
limited to the [`repo:status`] authorization scope (so this token can't
be used to write to anything other than commit statuses) that I've added
to the build environment.

[personal access token]: https://help.github.com/articles/creating-an-access-token-for-command-line-use/
[`repo:status`]: https://developer.github.com/v3/oauth/#scopes

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -73,7 +73,7 @@ test:
                  "context": "ci/many-worlds",
                  "state": "success",
                  "description": "Try the tests on the Many-Worlds build of this commit:",
-                 "target_url": "http://many-worlds.gomix.me/mathquill/mathquill/commit/'$CIRCLE_SHA1'/test/"
+                 "target_url": "http://many-worlds.glitch.me/mathquill/mathquill/commit/'$CIRCLE_SHA1'/test/"
                }'
 
     # Safari on Sauce can only connect to port 3000, 4000, 7000, or 8000. Edge needs port 7000 or 8000.
--- a/circle.yml
+++ b/circle.yml
@@ -65,6 +65,17 @@ dependencies:
 
 test:
   pre:
+    - |-
+      # Generate link to Many-Worlds build and add to GitHub Commit Status
+      curl -i -X POST https://api.github.com/repos/mathquill/mathquill/statuses/$CIRCLE_SHA1 \
+           -u MathQuillBot:$GITHUB_STATUS_API_KEY \
+           -d '{
+                 "context": "ci/many-worlds",
+                 "state": "success",
+                 "description": "Try the tests on the Many-Worlds build of this commit:",
+                 "target_url": "http://many-worlds.gomix.me/mathquill/mathquill/commit/'$CIRCLE_SHA1'/test/"
+               }'
+
     # Safari on Sauce can only connect to port 3000, 4000, 7000, or 8000. Edge needs port 7000 or 8000.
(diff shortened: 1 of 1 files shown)


laughinghan:
This will of course only effect PRs after this is merged into master, and only if the PR is branched off master after the commit that merges this in. It sure would be nice to have Many-Worlds links on all the old PRs, too. The tips, at least.

Many-Worlds currently only works if `make test` works, which was introduced in 11cb046a468020298b2c8b48ab590adf4118d20c (&ldquo;unit tests!&rdquo;). Finding the tip commits of every PR since 11cb046a468020298b2c8b48ab590adf4118d20c is a job for the Git command-line, so it's convenient that [the tips of PRs are all available as refs](https://help.github.com/articles/checking-out-pull-requests-locally/) of the form `refs/pull/123/head` in the GitHub remote repo:

```sh
$ git fetch origin refs/pull/*/head:refs/pull/origin/*
From github.com:mathquill/mathquill
 * [new ref]         refs/pull/100/head -> refs/pull/origin/100
 * [new ref]         refs/pull/101/head -> refs/pull/origin/101
 * [new ref]         refs/pull/102/head -> refs/pull/origin/102
...
```

We can then use [`git for-each-ref --contains <commit>`](http://stackoverflow.com/a/32987802/362030) to find the PR refs containing the commit that introduced `make test`, and pipe that to `xargs` to do the same `curl` request this PR does to each of those commits:

```sh

git for-each-ref refs/pull/origin/* \
    --contains 11cb046a468020298b2c8b48ab590adf4118d20c \
    --format '%(objectname)' \
| xargs -I {} \
    curl -i -X POST https://api.github.com/repos/mathquill/mathquill/statuses/{} \
         -u MathQuillBot:$GITHUB_STATUS_API_KEY \
         -d '{"context":"ci/many-worlds","state": "success","description":"Try the tests on the Many-Worlds build of this commit:","target_url": "http://many-worlds.gomix.me/mathquill/mathquill/commit/{}/test/"}'
```

(Note that because of `xargs`&rsquo; crazy 255-character limit on each argument that `-I` does a substitute in, the JSON can't be pretty-formatted like in this PR.)

(I tested this before running it by first piping to `xargs -I {} echo curl ...` and manually inspecting the first few `curl` commands that would be run. That&rsquo;s how I found out we were running afoul of the 255-char limit.)

Issue #727, opened 2017-03-06
Mac OS Chrome caret input bug


bheden:
In Chrome on Mac OS (Chrome 56, Mac OS Sierra 10.12.3) the exponent is not output, unless typed in twice. It appears to be an issue with Chrome reporting that a selection exists when typing certain characters, for instance ^ or ´.

The problem is present in the example field on http://mathquill.com/ so it's not specific for our implementation.

I made a dirty fix in our project by simply checking that the textarea value !== '^' in the hasSelection function. A real solution would require either distinguishing this selection in Chrome from the ones the function is meant to catch, or listing all characters that cause the problem and checking against that list rather than just the ^.

kschaefe:
If the field is empty, then inserting the ^ will seemingly do nothing.  This is because the ^ has nothing to typeset against.  Setting `supSubsRequireOperand` to `true` will only allow the use of ^ when there is a preceding value that can be typeset against.

I raised a similar problem and started working on a solution, but I've gotten stuck. Some of my work for that solution can be found in #467.

MathQuillBot:
**laughinghan** (via Slack):

@bheden: What language is your keyboard set to in your OS? Is it a language like French or Spanish where ^ and ` are used in digraphs to type accented characters like â and è? Does the problem go away if you set the language to English?

bheden:
@kschaefe: This problem occurs when typing simple exponentation, i.e. 10^4. I actually have to type 10^44 to have Mathquill print the four (or, to be precise, I have to press a key that triggers a keypress event after inputing the first 4). 

@MathQuillBot: I use swedish keyboard and the problem does go away when switching to english keyboard. Kind of weird since we don't use accented characters, at least not with ` or ^. French keyboard has the same issue. On the spanish I', afraid couldn't find the char. Not a swedish only problem at least ...

bheden:
Some more information:

When using swedish keyboard the problem is present in the latest versions of Opera, Firefox and Chrome in Mac OS. It's not present in Mac OS - Safari, and not in any browser in Windows 10. 

Opera and Chrome has the same explanation - `hasSelection` for the textarea returns true when characters that could be used for accented characters are input, i.e. ^ or `. 

[...]

PR #728, opened 2017-03-08
Fix webkit selection on input bug

bheden:
Webkit browsers in Mac OS sets a selection in the textarea when inputing certain characters using some keyboard layouts, for instance a caret sign ^ on a swedish keyboard. Allowing the typedText function to run for input events despite a selection being present should resolve the problem.


laughinghan:
Wow, fascinating, this fixes it? Can you think of any way to add an automated test for #727 to `saneKeyboardEvents.test.js`?

One thing I'm worried about: you confirmed this only happens with certain non-English keyboards. I believe that on those keyboards, ^ is a special character meant to be used in a digraph to type stuff like ê. With this change, is it impossible to type stuff like ê in MathQuill, since typing the ^ immediately results in a superscript? Was it already impossible to type?

bheden:
Trying to write ê with a french keyboard is buggy:

skarmavbild 2017-03-14 kl 08 32 14

But I'm afraid you're right about this change making it worse since the digraph is now not shown at all, rather than being wrongly positioned. 

I'll see if I can find another solution that solves the problem in both Webkit and Firefox without messing up the digraphs.

bheden:
I was a bit too quick in giving up on this ...

There are a few problems with circumflex characters (like ê) in MathQuill:

1. Exponantation mode is entered as soon as the user inputs a caret sign. 
2. The regexp `ch.match(/^[a-eg-zA-Z]$/)` is used to find letters. Since ê doesn't match that regexp, it is printed as a `VanillaSymbol` instead of `Letter`. Symbols are shown in the "Symbola" font which doesn't support this character. That, in turn, leads to the presentation seen in the screenshot in the previous comment.
3. Starting the input with a caret sign is prevented.

Since circumflex chars aren't really working anyway it's probably better to focus on exponantation.

bheden:
Changed to different solution. Circumflex now works (or rather doesn't work) the same way as before while exponentiation has been fixed. Someone with better insight in the code should be able to tell if I'm missing some case where the regexp will match but slicing the text causes problems.

kikoramathquill:
Did you check in Firefox?

bheden:
2^a in Firefox gives exponentiation. It is however not possible to write â in Firefox since exponentation mode is entered as soon as ^ is input.

PR #743, opened 2017-04-12
Escape latex in staticmath selectable span

eoghanmcilwaine:
Hey @laughinghan @stufflebear - here's a fix for a simple escaping issue we noticed.

ctrlr.exportLatex() was being directly parsed as html, causing issues with chars that are acceptable in latex but not in html.


laughinghan:
Beautiful, thanks @eoghanmcilwaine!

PR #744, opened 2017-04-19
On CI, cached Sauce Connect is outdated

laughinghan:
Circle builds 407-410 have all failed with:
```
{"error": "Your Sauce Connect version (4.3) is no longer supported. Please download the
latest version from https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy."}
```

e.g. https://407-598343-gh.circle-artifacts.com/0/tmp/circle-artifacts.TKNsK04/stdout_SauceConnect_12869.txt

(on https://circleci.com/gh/mathquill/mathquill/407#artifacts )

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -44,19 +44,27 @@ dependencies:
 
         mkdir -p ~/sauce-connect
         cd ~/sauce-connect
+
         if [ -x sc-*-linux/bin/sc ]; then
           echo Using cached sc-*-linux/bin/sc
         else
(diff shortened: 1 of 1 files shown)


laughinghan:
Sauce Connect also randomly fails quite often (https://git.io/vPN8v),
but those random failures take like 2 minutes, much slower than
redownloading, so it doesn't save any time to retry before trying
redownloading.

PR #747, opened 2017-05-09
Allow extending mq- class prefix at build time

jwmerrill:
Based on an env variable: MQ_CLASS_PREFIX

Note that this just does a dumb sed replace on `mq-`. This seems to be fine, but could be brittle in the future. An alternative would be to go through and change out usages of `mq-` with something that is more obviously meant to be swapped out by a preprocessor, but not sure it's worth it. Luckily, `mq-` is fairly unlikely to appear in other contexts.


jwmerrill:
Motivation for this is to allow embedding a widget that uses mathquill (like the Desmos calculator) onto a page that also uses mathquill for other purposes, and that may attach its own styles to `.mq-` classes.

laughinghan:
Why not just postprocess MathQuill's built files with `sed s/mq-/whatever-/g path/to/mathquill/*.{css,js}`?

As a bonus, if you know the platform on which your postprocessing will take place, you can use your platform's `sed` syntax for a word boundary, for extra safety (GNU: `sed s/\<mq-/whatever-/g`, BSD: `sed s/[[:<:]]mq-/whatever-/g`; meanwhile in Perl and JS, word boundaries are `\b`).

laughinghan:
Gonna close for now, feel free to re-open if you still think this should be merged.

PR #748, opened 2017-05-19
Added missing wrapper call!

Venryx:
The source code in this file was missing the
```
})();
```
at the end, to have the wrapper-function actually operate. (it seems the `function() {` line is just ignored otherwise)

### DIFF
--- a/src/intro.js
+++ b/src/intro.js
@@ -117,3 +117,5 @@ function bind(cons /*, args... */) {
 function pray(message, cond) {
   if (!cond) throw new Error('prayer failed: '+message);
 }
+
+})();


laughinghan:
It's not ignored and this is intentional. The files in `src` are concatenated together with `intro.js` at the beginning and `outro.js` at the end, so that the entire contents of `src` are wrapped in an [IIFE](http://benalman.com/news/2010/11/immediately-invoked-function-expression/).

Venryx:
You may do as you prefer, of course; but it is an odd/non-standard way of writing javascript files (they usually are standalone, valid javascript), and may confuse other potential contributors.

laughinghan:
It's not a preference, it was standard-ish at the time MathQuill was created (at least, it's what jQuery did). For a bit I was thinking of using Rollup, but now I'm actually thinking we should just switch to TypeScript.

In any case, it's certainly outdated and needs to be fixed. (This patch doesn't do that, though.)

PR #751, opened 2017-05-29
Fix #548, add a config option to StaticMath, making it possible to turn off mouse interaction

erikkrasner:
This PR is an update to #548.

Adds mouseEvents config option to StaticMath, which defaults to true. For a StaticMath element with MouseEvents set to false, no mouse events are bound.

This PR includes a demo example and some visual tests, and also corrects an inaccuracy in the demo copy that we considered too minor to be worth its own separate issue.

### DIFF
--- a/test/visual.html
+++ b/test/visual.html
@@ -163,6 +163,10 @@ $(function() {
 
 <p>Even in IE&lt;9, the background color of the parens and square root radical should be the background color of the selection.
 
+<p>Static math with mouseEvents set to false should not interact with the mouse: <span class="static-math-no-mouse-events">12 + 34</span></p>
+
+<p>Even in the case where it has an empty element: <span class="static-math-no-mouse-events">\sqrt{} </span></p>
+
 <h3>Dynamic mathquill-ification</h3>
 <table id="dynamic-initial">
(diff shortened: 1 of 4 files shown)


test/visual.html
```diff
@@ -163,6 +163,10 @@ <h3>Selection Tests</h3>
 
 <p>Even in IE&lt;9, the background color of the parens and square root radical should be the background color of the selection.
 
+<p>Static math with mouseEvents set to false should not interact with the mouse: <span class="static-math-no-mouse-events">12 + 34</span></p>
+
+<p>Even in the case where it has an empty element: <span class="static-math-no-mouse-events">\sqrt{} </span></p>
+
```
kschaefe: Can you add an inert field with an embedded MathQuillMathField?  I'd like to have that in the visual check to make it easy to ensure that embedded fields are not broken during future changes.

kschaefe: So, I added this locally and discovered that embedded MathQuillMathFields just don't work.  That's not a stopper from adding this functionality, since it's new and opt-in.  However, I think we should make effort to allow that long-term, so I'll add a new task for doing so.

PR #754, opened 2017-06-17
Added missing omicron

jlchereau:


jlchereau:
Sorry, I just realized that \omicron is an exception in LateX and a plain `o` should be used instead.

PR #760, opened 2017-07-10
Update Getting_Started.md

makrandgupta:


laughinghan:
@makrandgupta is MathQuill broken when using jQuery 1.11? It's supposed to support 1.5.2 **and above**, after all

makrandgupta:
I cant remember exactly what the error was, but MathQuill was acting up with older jQuery versions. Giving it the latest one fixed it. Thought I might as well update the public docs to use the supported version :)

laughinghan:
I don't think we want to recommend people use old versions of jQuery, newer versions are generally faster and often fix security faults. Thanks for your help though, let me know if you have any more information about the error with certain versions of jQuery.

makrandgupta:
@laughinghan I think you misunderstood the purpose of the pull request. I updated jQuery **from an older version to the minimum supported** version. Specifically from `1.11` to `1.52`

laughinghan:
I think you misread. You changed it from `1.11.0` to `1.5.2`. That's a change from `1.11.x` to `1.5.x`, which is from newer to older.

makrandgupta:
Oh wow, that's embarrassing. Sorry about that 😅

PR #761, opened 2017-07-11
allow pound symbol

relvao:


laughinghan:
Neither of `\pound` nor `\pounds` appear to be real LaTeX commands? http://mathb.in/151899

relvao:
Mathjax does not support it but it seems to be a valid symbol according to
Wikipedia
https://en.m.wikibooks.org/wiki/LaTeX/Special_Characters

We use katex, an alternative to mathjax that supports it.

I can no longer remember where I saw pound but I agree that it doesn't seem
to be valid. Will change that as soon as I'm in front of a computer.

PR #773, opened 2017-08-24
Update advancedSymbols.js to include integrals

CoryGH:
This includes integral symbols in addition to the preexisting \int and \oint [symbols](https://en.wikipedia.org/wiki/Integral_symbol#Extensions_of_the_symbol):

- \iint

- \iiint

- \varointclockwise

- \ointctrclockwise

- \oiint

- \oiiint

In addition it includes two symbols not covered by latex but following the same nomenclature:

- \intclockwise

- \intctrclockwise

### DIFF
--- a/src/commands/math/advancedSymbols.js
+++ b/src/commands/math/advancedSymbols.js
@@ -201,6 +201,14 @@ LatexCmds.square = bind(VanillaSymbol, '\\square ', '&#11036;');
 
 //variable-sized
 LatexCmds.oint = bind(VanillaSymbol, '\\oint ', '&#8750;');
+LatexCmds.iint = bind(VanillaSymbol, '\\iint ', '&#8748;');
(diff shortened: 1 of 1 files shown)


CoryGH:
@kschaefe Where would the best place for these be?  I see a few `\int ` entries but no `\oint ` entries, would you like me to add the `\oint ` ones as well?

kschaefe:
Yes.  Please have something for each new items you added.  Makes it very easy to validate the look across multiple browsers.

If there are related symbols, it's best to group them, otherwise you can make a new little section in there.

CoryGH:
@kschaefe I'll try to get to this by the end of the weekend if possible (probably Monday.)

CoryGH:
A separate pull request has been created which includes visuals for all integral symbols supported: https://github.com/mathquill/mathquill/pull/777

PR #777, opened 2017-09-16
includes integral tests

CoryGH:
Includes tests requested in https://github.com/mathquill/mathquill/pull/773


CoryGH:
I don't have time to make the changes in the foreseeable future and it works on my end.  If you'd like the symbols added feel free to use the code I submitted.

PR #781, opened 2017-10-11
Ia1 295 fix issue with sqrt when zoomed

asgeir-inspera:


kschaefe (changes requested):
There is a lot of unrelated cruft in the commits here.  Looks like local changes that don't need to be part of the main repo.  If they do, we should target them in smaller more isolated chunks.  There are a lot of conflicts here that need to be resolved.

I'd recommend ensuring that the changes for the sqrt fix are isolated from other changes.  Merge master back into you patch and resolve any outstanding conflicts.

It may be best to close this pull request and open a new one, cherry-picking over just the necessary changes.

Also, you'll need to sign the contributor agreement before we can merge and work that you give us.

PR #782, opened 2017-10-19
Improve MathElement.text() compute speed

sclower:
Not sure how often this is used, but we discovered a subtle issue with .text() which was causing some significant performance problems with deeply nested constructs.


jwmerrill:
The issue here is that `.text()` was calling itself recursively twice instead of just once, so the runtime was exponential in the depth of the tree instead of linear in the number of nodes.

PR #787, opened 2017-11-30
Fixing whitespace parsing bug for \langle,\rangle

dominique-unruh:
`\left\langle1235\right\rangle` is not parsed by MathQuill (observed by https://github.com/mathquill/mathquill/pull/659#issuecomment-348105353), but `\left\langle 1235\right\rangle` is. The reason is that `\b` matches the end of a word only, and according to Javascript's regular expression semantics, `langle12345` is one word. (While for LaTeX, 12345 cannot be part of a word.)

But we cannot just remove `\b` because it was introduced to avoid confusion with, e.g., `\langlerfish` (see above).

To fix this, we replace every `\b` by `(?![a-zA-Z])` (this is a negative lookahead that only matches when no letter follows). Now things seem to work.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -747,14 +747,14 @@ LatexCmds.left = P(MathCommand, function(_) {
     var succeed = Parser.succeed;
     var optWhitespace = Parser.optWhitespace;
 
-    return optWhitespace.then(regex(/^(?:[([|]|\\\{|\\langle\b|\\lVert\b)/))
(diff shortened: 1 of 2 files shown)


dominique-unruh:
The ci/circleci tests fail, but on my local build all seems to work. Can somehow tell me how to interpret the circleci output? I can't tell what needs to be fixed...

kschaefe:
Thanks for the quick updates!

PR #802, opened 2018-03-28
Lessc 3.0.1 breaks build

electriccode:
According to https://github.com/less/less.js/issues/2911 `@var:;` is non-standard and `@var:~'';` is the preferred approach.


electriccode:
Rejected, reason being, I was unable to sign the CLA as the commit username was different in this PR.

Issue #797, opened 2018-03-07
lessc 3.0.1 breaks build


charlesverge:
lessc 3.0.1 returns the following error
./node_modules/.bin/lessc  ./src/css/main.less > build/mathquill.css
undefinedError: Cannot read property 'eval' of null

Using less 2.7.2 fixes this issue.

kschaefe:
Fixed by #803.

PR #803, opened 2018-03-28
#797 less 3.0.1 breaks build because @var:; is non standard

electriccode:
https://github.com/less/less.js/issues/2911


electriccode:
Was unable to sign the CLA as the commit username was different in this PR.

laughinghan:
Oh brilliant, this appears to be why http://many-worlds.glitch.me/mathquill/mathquill/branch/master/ is broken

PR #805, opened 2018-03-28
onRenderComplete callback for MathQuill.StaticMath

electriccode:


electriccode:
Can somebody check the circleci build? Looks like its a timeout issue

PR #852, opened 2019-03-31
Match example to proposal: use jQuery 1.5.2

LucasLarson:
We say that “[MathQuill depends on jQuery 1.5.2+](https://github.com/mathquill/mathquill/blob/4e2a50f7dd54df06bcc7fc003c6df8cc87491224/docs/Getting_Started.md#download-and-load)” but we link to 1.11.0. I’m proposing our example link to 1.5.2 (which is [what we do in quickstart.html](https://github.com/mathquill/mathquill/blob/7f7f0af6f7f1a4591bce8b257a539bc12a5a1dda/quickstart.html#L14)).


LucasLarson:
+ confirmed explicit 1.5.2 in https://github.com/mathquill/mathquill/commit/ead472c3c3b835a1ed21d0ac11611fae195dfdb8

PR #877, opened 2019-10-20
Update README.md

151871n:


laughinghan:
Is this an accident? Seems fine how it is.

PR #893, opened 2020-05-08
Add parentheses to superscript text template

### LINKED ISSUE #840: can't get correct text from MathField
If i have a editable MathQuill and i will write in it " 1 / 2 " and then get content:
`mathField.latex() --> \frac{1}{2}`,
`mathField.text() --> (1)/(2)` (where did the brackets come from?)
But if i will write "2 ^ -3":
`mathField.latex() --> 2^{-3}`
`mathField.text() --> 2^-3` (why without brackets?)

As a result if i will write "2 ^ 1 / 2" in field i can't get valid text like "2 ^ (1 / 2)" or "2 ^ ((1) / (2))", i get "2 ^ (1) / (2)" and parse it how `2 ^ 1 / 2 = 1`. 

Maybe a can paste brackets automatically when user write " ^ "? And if you automatically write brackets in the fractions, why you can't write it in the powers? Thanks

sashaalesin:
Fix #840

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -250,6 +250,13 @@ var SupSub = P(MathCommand, function(_, super_) {
     }
     return latex('_', this.sub) + latex('^', this.sup);
   };
+  _.text = function() {
+    function text(prefix, block) {
+      var l = block && block.text();
+      return block ? prefix + (l.length === 1 ? l : '(' + (l || ' ') + ')') : '';
+    }
+    return text('_', this.sub) + text('^', this.sup);
+  };
   _.addBlock = function(block) {
     if (this.supsub === 'sub') {
       this.sup = this.upInto = this.sub.upOutOf = block;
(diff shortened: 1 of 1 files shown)


sashaalesin:
image
Before:
`.text()` -> "c_12"
After:
`.text()` -> "c_1^2"

src/commands/math/commands.js
```diff
@@ -346,7 +353,7 @@ LatexCmds['^'] = P(SupSub, function(_, super_) {
     +   '<span class="mq-sup">&0</span>'
     + '</span>'
   ;
-  _.textTemplate = [ '^' ];
+  _.textTemplate = ['^(', ')'];
```
laughinghan, line 356: I think this is ignored if you override by providing a `.text()`

laughinghan (approved):

laughinghan:
Great work! Tests would also be nice

Issue #460, opened 2015-08-20
Feature Request: Unwrap parens when block becomes the numerator of a fraction


jwmerrill:
or when other operations are applied to it as a block, like taking the square root.

laughinghan:
Duplicate of #460

Issue #401, opened 2015-05-20
Unwrap parens when converting an equation into a fraction?


eluberoff:
A relatively common way to make a multi-part fraction is to wrap the numerator in parentheses and then divide. In an old equation editor I built, we'd peel off the parens in this circumstance. So typing (1+2)/3 would result in

```
1+2 
-----
  3
```

Premis was that that's what you'd want in the vast majority of cases that you use parens and then divide. Related to discussion in #396

laughinghan:
I actually love this. I wonder if @jneen has an opinion on this?

laughinghan:
@jwmerrill @eluberoff We got a PR from Professor Lippman. Do you think this is disruptive enough to need a flag? Or people probably won't even notice?

eluberoff:
Such a good question. I think in most cases this might just expected behavior and folks won't notice, but there's now a bit of extra complexity, at least in Desmos as a proxy. The case I'm worried about is that you can do scalar operations on points, so like this is supported: https://www.desmos.com/calculator/1wzcgykpya

and stripping parens in that case would be unexpected behavior. I guess that might argue for a flag? Low confidence.

PR #894, opened 2020-05-18
Unwrap the parens around the numerator when forming a fraction

drlippman:
So typing (2+3)/4 would display without the parens around the numerator

This is a possible solution to #401 and #460.

I intentionally limited it to blocks wrapped in round paren brackets to avoid breaking potentially legitimate cases of a bracketed quantity in a fraction.

I'm not sure whether this should be hidden under a config flag.

### DIFF
--- a/src/commands/math/commands.js
+++ b/src/commands/math/commands.js
@@ -485,6 +485,16 @@ CharCmds['/'] = P(Fraction, function(_, super_) {
         cursor[L] = leftward;
       }
     }
+    if (this.replacedFragment) {
+      if (this.replacedFragment.ends[L] instanceof Bracket &&
+        this.replacedFragment.ends[L].sides[L].ctrlSeq == '(' &&
+        this.replacedFragment.ends[L].sides[R].ctrlSeq == ')'
+      ) {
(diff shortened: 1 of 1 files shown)


src/commands/math/commands.js
```diff
@@ -485,6 +485,16 @@ CharCmds['/'] = P(Fraction, function(_, super_) {
         cursor[L] = leftward;
       }
     }
+    if (this.replacedFragment) {
+      if (this.replacedFragment.ends[L] instanceof Bracket &&
```
laughinghan, line 489: Will this delete the `+c` if I'm here and I press `/`?
Screen Shot 2020-09-14 at 20 31 18


```suggestion
      if (this.replacedFragment.ends[L] instanceof Bracket &&
        this.replacedFragment.ends[L] === this.replacedFragment.ends[R] &&
```

laughinghan (changes requested):

laughinghan:
Nice work! And yeah, great call on restricting this to `( ... )`-only.

src/commands/math/commands.js
drlippman: Good catch.

PR #902, opened 2020-07-10
Feature make static

kschaefe:
Add the ability to make inner math fields static and the ability to change them back to editable fields.


kschaefe:
`acpiplica` does not need to sign the CLA.  We own her work product for this, my `kschaefe` signature should be enough.

manstie:
Was looking for this feature to convert between editable and static.

laughinghan (approved):

PR #911, opened 2020-09-14
Fix pasting into text blocks

jaltekruse:
- previously it would insert a math node into a text block, if you
  happened to paste in latex
- if you pasted in somthing that wasn't latex you would get mathquill
  source in the latex output
- Update test framework to print both the generic message like a failure
  when comparing two values as well as the cutom message priovided if
  there is one


laughinghan (approved):
Dooope

PR #914, opened 2020-09-22
Feature/improve docs

kschaefe:


kschaefe:
Hmm, must have gotten the history wrong.  Only one commit should have been there.

PR #919, opened 2020-11-05
Added support for some new symbols

IsmailAlamKhan:
I added support for some new symbols for using them with LaTeX values

### DIFF
--- a/src/commands/math/basicSymbols.js
+++ b/src/commands/math/basicSymbols.js
@@ -2,14 +2,14 @@
  * Symbols for Basic Mathematics
  ********************************/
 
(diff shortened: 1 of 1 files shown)


shariful2011 (approved):

kevinbarabash:
It's nice seeing more symbols being added.  Does the font (Symbola) support all of these new symbols?  It would be nice to see screenshots of MathQuill displaying all of these new symbols.  One of the issue I can see with this PR moving forward is that there's a bunch of extraneous formatting changes.

It would be nice if MathQuill supported a way to add symbol definitions via the config options so that people could just add whatever symbols they needed themselves.

Issue #925, opened 2021-03-08
MacOS build failed:sed: 1: "build/mathquill-basic.js": undefined label 'uild/mathquill-basic.js'


rawbin-:
using master and run npm install on MacOS 11.2.2

# make basic 

cat src/intro.js node_modules/pjs/src/p.js src/tree.js src/cursor.js src/controller.js src/publicapi.js src/services/parser.util.js src/services/saneKeyboardEvents.util.js src/services/exportText.js src/services/focusBlur.js src/services/keystroke.js src/services/latex.js src/services/mouse.js src/services/scrollHoriz.js src/services/textarea.js src/commands/math.js src/commands/math/basicSymbols.js src/commands/math/commands.js src/outro.js build/.exists--used_by_Makefile | ./script/escape-non-ascii > build/mathquill-basic.js
sed -i     s/{VERSION}/v0.10.1/ build/mathquill-basic.js
sed: 1: "build/mathquill-basic.js": undefined label 'uild/mathquill-basic.js'
make: *** [build/mathquill-basic.js] Error 1

jaltekruse:
It looks like this might be part of the issue the 'b' looks to have been chopped off of the from of this path -  undefined label 'uild/mathquill-basic.js'.

Are you sure you have a clean checkout of master? Could you post your npm version?

rawbin-:
![image](https://user-images.githubusercontent.com/8169783/115646295-77f41f80-a354-11eb-8dba-e91bf4dae98b.png)

this line makes it to work

jwmerrill:
I just upgraded from macOS Catalina to Big Sur, and I'm running into this now too.

I see that the `sed` that macOS is shipping now supports the `-r` option (excerpt from `man sed`):
```
-r      Same as -E for compatibility with GNU sed.
```

So this logic is now detecting macOS's sed as "GNU", even though it still needs the "BSD" behavior for `-i`.

jwmerrill:
A bit of googling is telling me that using `perl` for this purpose is more portable than using `sed`.

I tried doing
```
SED_IN_PLACE = perl -pi -e
```

and this gets the build working for me again, although I think it will be better to just inline this to avoid confusingly calling it SED now that it's perl.

I can put up a PR for this.

PR #932, opened 2021-05-09
Fix Big Sur build

jwmerrill:
The build system has been using `sed -i` to do inline string replacement in files, but the `-i` option to sed is not POSIX and its behavior varies between GNU and BSD.

The build system attempted to detect which system you were on by checking the behavior of a different flag (`-r`), but this detection no longer works correctly on macOS Big Sur because it has added the `-r` flag for compatibility with GNU (causing it to be incorrectly detected as GNU).

Fix this by replacing `sed -i` with `perl -pi -e`. Perl is very widely available, and its behavior here is more standard across systems.

Ref:
* https://stackoverflow.com/questions/3156178/how-do-i-use-perl-like-sed/37686753#37686753
* https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux/22247781#22247781

Fixes https://github.com/mathquill/mathquill/issues/925

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,14 +15,6 @@ ifneq ($(shell node -e 'console.log("I am Node.js")'), I am Node.js)
   $(error Please install Node.js: https://nodejs.org/ )
 endif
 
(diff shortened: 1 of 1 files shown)


laughinghan (approved):
Love it! Great work!

PR #934, opened 2021-06-01
Fix redundant focus event on paste

jwmerrill:
Mathquill has a workaround for Linux that makes sure the appropriate element is focused on a paste event. This focus event bubbles in the usual way, so it may be observed outside mathquill.

This extra focus event started tickling a bug in the Desmos graphing calculator after upgrading from jQuery 2.x to 3.x. I suspect there may be a difference between these versions (possibly only in some browsers) in handling the case that the appropriate element already had focus.

Switch to only triggering this event if the appropriate element is not already focused to reduce the chance of triggering focus related application bugs.


kschaefe (approved):
Simple, appropriate change.  Looks good.

PR #937, opened 2021-07-29
Fix #936 ("\class{}{} command not translating back to LaTeX")

### LINKED ISSUE #936: \class{}{} command not translating back to LaTeX
The `\class{}{}` command, used for giving CSS classes to its content, doesn't translate back to LaTeX, meaning that `.latex()` will return garbage and not the original command.

Example that can be run on http://mathquill.com:

```
var math = MathQuill.getInterface(2)($("#math-field")[0]);
math.latex("\\class{test}{abc}");
alert(math.latex());
```

This will alert "`class{abc}`", swallowing the backslash and the "`{test}`".

TomasRiker:
This fixes issue #936 by giving the `Class` parser a `latex` method, so that it can output its own LaTeX representation, just like `TextColor`.


TomasRiker:
Note: The failed test on CircleCI is not due to the code. Something is broken in the build process.

PR #949, opened 2021-11-18
Upgrade SauceConnect from 4.5.4 to 4.7.1

jwmerrill:
Intended to fix CI failures like the one in https://github.com/mathquill/mathquill/pull/948

Error msg:
> error querying from https://saucelabs.com/rest/v1/*********/tunnels, error was: 400: Your Sauce Connect version (4.5.4) is no longer supported (must be newer than 4.5.4). Please download the latest version from https://docs.saucelabs.com/secure-connections/sauce-connect/installation (ref: 84650008). HTTP status: 400 Bad Request

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -90,8 +90,8 @@ jobs:
           if [ -x sc-*-linux/bin/sc ]; then
             echo Using cached sc-*-linux/bin/sc
           else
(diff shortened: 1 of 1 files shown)


jwmerrill:
CI is passing now, so it appears this fixes the problem

jaltekruse (approved):

Issue #955, opened 2021-12-16
Mathquill allows pasting UTF-8 characters/symbols but does not convert them to Latex in output


jaltekruse:
This breaks compatibility with other tools like Katex.

This is my current (incomplete) workaround for the issue in https://freemathapp.org

https://github.com/jaltekruse/Free-Math/blob/3bae6e16ed8f698a22660755a5195fa75e57d870/src/TeacherInteractiveGrader.js#L1229

jaltekruse:
Figured out there is a feature in the code to handle this, we just need to add more symbols.

For example, all symbols that have a version that includes the UTF-8 character stored into the LatexCmds object work just fine for the case of pasting in the character, it gets immediately converted into the standardized latex in the output.

from advancedSymbols.js

LatexCmds['≈'] = LatexCmds.asymp = LatexCmds.approx = bind(BinaryOperator,'\\approx   ','&asymp;');

jaltekruse:
Might be easiest to just do a lookup into this list of all of the HTML entities, which is how many of the symbols are specified in MathQuill.

https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references

jaltekruse:
Here is the list of symbols I found declared in MathQuill that I paste in the repro video.

⬜∐√→∉≅≡⊕⊗≠∗∴∵∝≈∈∋∌⊂⊃⊄⊅⊆⊇⊈⊉ℕℙℤℚℝℂℍ◇△⊖⊎▽⊓⊲⊔⊳⊙◯≀⊨≺≻≼≽

≃∣≪≫∥∦⋈⊏⊐⌣⊑⊒≐⌢⊦⊣≮≯←⇐⇒↔↕⇔⇕↦↗↩↪↘↼⇀↙↽⇁↖…⋯⋮⋱ℓ⊤♭♮♯℘⊥♣♢♡♠▱

∮∩∪∨∧⌊⌋⌈⌉∇ℏÅ∘•∖¬↓⇓↑⇑ℜℑ∂∞£ℵ∃∄∅°∠∡

′″&αβγδζηθικμνξρστχψωϕφϵεϖςϑυϝϰϱπλϒΓΔΘΛΞΠΣΦΨΩ∀−±∓·<>≤≥×÷

∉≅≡⊕⊗αβγδζηθικμνξρστχψωΓΔΘΛΞΠΣΦΨΩ∀

PR #960, opened 2022-01-07
Fix pasting UTF-8 symbols

jaltekruse:
Mathquill allows pasting UTF-8 characters/symbols but does not convert them to Latex in output.

This breaks compatibility with other tools like Katex.

https://user-images.githubusercontent.com/365366/148617104-08342721-c97c-4365-8dca-4419d11f73ae.mp4

https://user-images.githubusercontent.com/365366/148617105-59dfe27c-a90e-4ccb-901b-73a68265ac1e.mp4


jaltekruse:
I understand we likely want to prune the giant html entities reference I'm using to look up the resolved utf-8 character for a given symbol. It should be able to be worked into the build to generate a minimal list of just the data we need, but I probably would check in the whole list to the repo, so we don't need to hunt down more entries as symbols are added to MathQuill.

jaltekruse:
Here is the list of symbols I found declared in MathQuill that I paste in the repro video.

```
⬜∐√→∉≅≡⊕⊗≠∗∴∵∝≈∈∋∌⊂⊃⊄⊅⊆⊇⊈⊉ℕℙℤℚℝℂℍ◇△⊖⊎▽⊓⊲⊔⊳⊙◯��≀⊨≺≻≼≽

≃∣≪≫∥∦⋈⊏⊐⌣⊑⊒≐⌢⊦⊣≮≯←⇐⇒↔↕⇔⇕↦↗↩↪↘↼⇀↙↽⇁↖…⋯⋮⋱ℓ⊤♭♮♯℘⊥♣♢♡♠▱

∮∩∪∨∧⌊⌋⌈⌉∇ℏÅ∘•∖¬↓⇓↑⇑ℜℑ∂∞£ℵ∃∄∅°∠∡

′″&αβγδζηθικμνξρστχψωϕφϵεϖςϑυϝϰϱπλϒΓΔΘΛΞΠΣΦΨΩ∀−±∓·<>≤≥×÷

∉≅≡⊕⊗αβγδζηθικμνξρστχψωΓΔΘΛΞΠΣΦΨΩ∀
```

jaltekruse:
Looks like github has some trouble with some of the symbols, here's the raw 
[mathsymbols.txt](https://github.com/mathquill/mathquill/files/7832575/mathsymbols.txt)
file.

jaltekruse:
This fixes - https://github.com/mathquill/mathquill/issues/955

src/commands/math/basicSymbols.js
```diff
@@ -257,7 +257,7 @@ LatexCmds.f = P(Letter, function(_, super_) {
 LatexCmds[' '] = LatexCmds.space = bind(VanillaSymbol, '\\ ', '&nbsp;');
 
 LatexCmds["'"] = LatexCmds.prime = bind(VanillaSymbol, "'", '&prime;');
-LatexCmds['″'] = LatexCmds.dprime = bind(VanillaSymbol, '″', '&Prime;');
+LatexCmds['″'] = LatexCmds.dprime = bind(VanillaSymbol, "''", '&Prime;');
```
laughinghan, line 260: This won't round-trip right?

laughinghan (approved):
Let's include links to the source of the JSON as a comment in `htmlEntities.js`

src/commands/math/basicSymbols.js
```diff
@@ -526,3 +528,26 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
   bind(BinaryOperator,'\\div ','&divide;', '[/]');
 
 CharCmds['~'] = LatexCmds.sim = bind(BinaryOperator, '\\sim ', '~', '~');
+
+for (var cmd in LatexCmds) {
+  if (LatexCmds.hasOwnProperty(cmd)) {
+      constructor = LatexCmds[cmd];
+      var block = constructor();
```
laughinghan, line 535: this variable seems like it should be called command or symbol or something. Block in MathQuill specifically means a MathBlock object containing zero or more commands, whereas this is exactly one command object

src/commands/math/basicSymbols.js
```diff
@@ -526,3 +528,26 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
...
+      // find templates with a unicode character code
+      // we can programatically generate the character in a string
+      var findUnicodePattern = /&#(.*);/;
```
laughinghan, line 538: technically this is called a "numeric character reference" and the other one a "named character (entity) reference" but I'm not sure those are actually more legible lmao

https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references

src/commands/math/basicSymbols.js
```diff
@@ -526,3 +528,26 @@ LatexCmds['÷'] = LatexCmds.div = LatexCmds.divide = LatexCmds.divides =
...
+      var arr = findUnicodePattern.exec(block.htmlTemplate);
+      if (arr && arr[1]) {
+        LatexCmds[String.fromCharCode(parseInt(arr[1]))] = constructor;
+      } else {
+        var findHtmlEntityPattern = /&(.*);/;
+        var maybeEntity = findHtmlEntityPattern.exec(block.htmlTemplate);
+        if (maybeEntity && maybeEntity[1]) {
+          var entityInfoLookup = htmlEntities['&' + maybeEntity[1] + ';'];
```
laughinghan, line 546: wouldn't it be simpler to not bother with the capturing group in this case?

Issue #950, opened 2021-11-22
Superscript vertical height may not be adjusted after deleting characters


jwmerrill:
Repro:

1. In mathquill master (note, not mathquill.com currently because that is behind master at the moment), paste in `\left(x^{\frac{1}{\frac{2}{3}}}\right)^2`
    <img width="58" alt="Screen Shot 2021-11-22 at 3 15 09 PM" src="https://user-images.githubusercontent.com/48409/142928872-74702ee7-5136-4546-8c1e-ba8210cb322d.png">
2. Place your cursor after the `3` innermost denominator
    <img width="57" alt="Screen Shot 2021-11-22 at 3 16 09 PM" src="https://user-images.githubusercontent.com/48409/142929003-f28ebc56-11a1-4971-a10f-6a1286fb9b23.png">
3. Press backspace four times to delete both denominators
    <img width="51" alt="Screen Shot 2021-11-22 at 3 16 38 PM" src="https://user-images.githubusercontent.com/48409/142929070-81d3939c-43e6-4541-a17c-30a3ca2c5d12.png">
    
This leaves a superscript stranded far above the parens it applies to.

Note, you don't actually need a nested fraction--that just makes the issue easier to see.

Relevant issues:
* https://github.com/mathquill/mathquill/issues/763
* https://github.com/mathquill/mathquill/pull/764
* https://github.com/mathquill/mathquill/pull/779

I think this comment may be saying this is a known caveat of the current implementation: https://github.com/mathquill/mathquill/pull/779#issuecomment-332029992

laughinghan:
Yeah #764 and #779 don't really work, the SupSub doesn't get reflow-ed when its sibling changes in height

PR #947, opened 2021-11-17
Merge in the desmos fork

jwmerrill:
Note this is part of a chain of PRs  that are concurrently under review to incorporate extensive work done by the Desmos team. The PRs all target one another in series. There is one big changeset that starts everything and two reformatting/restructuring changesets later in the series, so this is being done to have cleaner diffs, but without creating tons of individual PRs.

We are intending to merge all of these PRs at the same time when we are done reviewing all of them. The largest bulk of the changes have been posted for a few months in #947.

There a small inconvenience looking through the branches, you will see a number of "merge in branch feature-name-xyz resolves #YYYY" These will link you to issues or PRs in the mathquill org, but they are actually referencing PR numbers in the desmos fork, to see the conversations on these issues, you can click the links and then just replace the org part of the url to be desmosinc instead of mathquill, i.e change `https://github.com/mathquill/mathquill/issues/168` to `https://github.com/desmosinc/mathquill/issues/168`.

The chain of PRs:
1. Incorporate the Demsos fork, changes from the past couple of years, adds screen reader support and many fixes and smaller features - https://github.com/mathquill/mathquill/pull/947

2. Convert the codebase to typescript
https://github.com/mathquill/mathquill/pull/967

3. A few various fixes after the typescript conversion
https://github.com/mathquill/mathquill/pull/968

4. Add Prettier to atuo-format the code
https://github.com/mathquill/mathquill/pull/969

5. Fix selection performance
https://github.com/mathquill/mathquill/pull/970

---------------------

Adds screen-reader support and other features, optimizations and fixes.

Adds many new config options to control behaviors that Desmos wants to opt in to:

- `typingAsteriskWritesTimesSymbol: boolean;`
    -  When `true`, `*` writes `\times` instead of `\cdot`
- `typingSlashWritesDivisionSymbol: boolean;`
    - When `true`, `/` writes `\div` instead of `\frac`
- `typingPercentWritesPercentOf: boolean;`
    - When `true`, `%` writes `\%\operatorname{of}` instead of `\%`
- `resetCursorOnBlur: boolean;`
    - When `true`, MathQuill no longer tries to remember the last cursor position after a blur
- `enableDigitGrouping: boolean;`
    - When `true`, adds a thin space as a place value separator to large numbers after a short debounced editing timeout
- `disableCopyPaste: boolean;`
    - When `true`, disables copy and paste operations
- `ignoreNextMouseDown: (el: MouseEvent) => boolean;`
[...]

PR #779, opened 2017-09-25
Fix for { fix for Superscript of tall expression (issue #763) }

savthegood:
Fix to reflow properly superscript in cases of combo of sub+sup. Move reflow into SupSub so case a_b,right,^x is reflow-ed too. Use margin-bottom to lift superscript and dont touch vertical-align of mq-subsup (so rendering of subs is not messed).

PR #764, opened 2017-07-17
 Fix for Superscript of tall expression (issue #763) #1 

savthegood:
Superscript of very tall expression is rendered kind of wrong, e.g. `(-5^(1/2))^2+3` (here 2 is superscript):
image

with fix it will be smth like:
image

PR #964, opened 2022-01-26
Merge in the desmos fork

jwmerrill:
Adds screen-reader support and other features, optimizations and fixes.

There are a couple large structural changes in here that touch pretty much every line in every file
* The codebase has been ported from JS to Typescript (see https://github.com/desmosinc/mathquill/pull/168 for just these changes)
* The codebase is now being formatted by prettier. There's a precommit hook to format staged files. (see https://github.com/desmosinc/mathquill/pull/182 for just these changes)

For review, it may be easiest to look at #947 for changes that landed before the TS port and the commits after 23d2a1b for changes that landed after the TS port.

In addition to the new options added in https://github.com/mathquill/mathquill/pull/947/, subsequent work added two new config options.
- `quietEmptyDelimiters: string;`
    - space-separated list of delimiters (such as parens, brackets, and curly braces) for which Mathquill will not display a gray box inside empty pairs. Desmos configures this as `quietEmptyDelimiters: '( ['` to avoid showing a gray box between empty parens and square brackets.
- `disableAutoSubstitutionInSubscripts: boolean;`
    - When `true`, `autoCommands` and `autoOperatorNames` are not substituted in subscripts. For example, if `pi` is an autocommand, typing `a` `_` `p` `i` would result in the subscript containing two letters instead of a single `π` symbol.


jwmerrill:
Copying the commit comment from 9ac1090 for more visibility because I think it's important:

> Fixes https://github.com/mathquill/mathquill/issues/950 by reverting https://github.com/mathquill/mathquill/pull/764 and https://github.com/mathquill/mathquill/pull/779
> 
> Unfortunately, this leaves us back in a place where superscript and subscript placement isn't very good when they're attached to tall blocks, but I think if we're going to fix that, we need to find a way to do it that's more robust to continuing to edit the expression.

sclower:
## New Config Options

Note: config.md has not been updated with all of the below options as Desmos uses some internally, and we may still find a need to change them before they are set in stone.

- ignoreNextMousedown: (_el:MouseEvent) => boolean;
- typingAsteriskWritesTimesSymbol?: boolean;
- typingSlashWritesDivisionSymbol: boolean;
- typingPercentWritesPercentOf?: boolean;
- resetCursorOnBlur?: boolean | undefined;
- enableDigitGrouping?: boolean; // adds visual spacing between digits
- disableCopyPaste?: boolean;
- onPaste?: () => void;
- onCut?: () => void;
- overrideTypedText?:(text:string) => void; // allows caller to intercept typed text before it is committed
- overrideKeystroke: (key:string, event:KeyboardEvent) => void; // allows caller to intercept keystroke before Mathquill's default handlers take over
- quietEmptyDelimiters: { [id:string]:any; }; // space-separated list of delimiters (such as parens, brackets, and curly braces) for which Mathquill will not provide a gray outline.
- disableAutoSubstitutionInSubscripts?: boolean; // allows caller to set whether auto commands, auto operators, and automatically parenthesized functions are expanded when typing into a subscript: e.g. should x_{pi} be changed to x_{\\pi}?

Makefile
```diff
@@ -106,34 +114,44 @@ css: $(BUILD_CSS)
...
 clean:
 	rm -rf $(BUILD_DIR)
-
-$(PJS_SRC): $(NODE_MODULES_INSTALLED)
+# This adds an entry to your local .git/config file that looks like this:
+# [include]
+# 	path = ../.gitconfig
+# that tells git to include the additional configuration specified inside the .gitconfig file that's checked in here.
+setup-gitconfig:
+	@git config --local include.path ../.gitconfig
```
jwmerrill, line 122: Note, running `make setup-gitconfig` once will make blame skip the commits where prettier was run, which helps a lot with preserving history.

We could consider documenting this somewhere more prominent, or trying to run this as part of install, or something like that.

jaltekruse:
We're closing this PR in favor of creating a chain of PRs that segment the different parts of this work that will have cleaner diffs to look at.

Issue #168, opened 2013-03-24
Clicking in LatexCommandInput focuses wrong thing


laughinghan:
because in 647aa0d4240bbe5339f1141b07c4fead768ad5d4 I didn't take into account the fact that `Cursor::insertAdjacent` violates the Law of Demeter and directly does `this.parent.jQ.addClass('hasCursor')`--why the fuck did I do that? Though I feel like this is just a symptom of `focus`/`blur` being thoroughly fucked, starting with its flawed conception which I'm pretty sure was my original attempt at adding `TextBlock` without special casing--apparently I thought that methods without contracts was better than special casing (instead of a contract, I just knew exactly when and where the method was called, which was when a cursor moved into a block via the keyboard but not when clicking).

PR #967, opened 2022-01-28
Convert codebase to typescript

jaltekruse:
For reference, the PR where this landed in the Desmos fork - https://github.com/desmosinc/mathquill/pull/168

---------
Note this is part of a chain of PRs  that are concurrently under review to incorporate extensive work done by the Desmos team. The PRs all target one another in series. There is one big changeset that starts everything and two reformatting/restructuring changesets later in the series, so this is being done to have cleaner diffs, but without creating tons of individual PRs.

We are intending to merge all of these PRs at the same time when we are done reviewing all of them. The largest bulk of the changes have been posted for a few months in #947.

There a small inconvenience looking through the branches, you will see a number of "merge in branch feature-name-xyz resolves #YYYY" These will link you to issues or PRs in the mathquill org, but they are actually referencing PR numbers in the desmos fork, to see the conversations on these issues, you can click the links and then just replace the org part of the url to be desmosinc instead of mathquill, i.e change `https://github.com/mathquill/mathquill/issues/168` to `https://github.com/desmosinc/mathquill/issues/168`.

The chain of PRs:
1. Incorporate the Demsos fork, changes from the past couple of years, adds screen reader support and many fixes and smaller features - https://github.com/mathquill/mathquill/pull/947

2. Convert the codebase to typescript
https://github.com/mathquill/mathquill/pull/967

3. A few various fixes after the typescript conversion
https://github.com/mathquill/mathquill/pull/968

4. Add Prettier to atuo-format the code
https://github.com/mathquill/mathquill/pull/969

5. Fix selection performance
https://github.com/mathquill/mathquill/pull/970

----------

The goal of this work was to port to typescript while changing as little structure and runtime behavior as possible. This helps preserve MQ's history and hopefully makes it less likely to introduce logic bugs in a PR that's admittedly hard to review. To achieve this, there are places that aren't perfectly idiomatic typescript. In particular, there are a lot of type casts in places where MQ's very dynamic structure makes it hard to get accurate type inference without changing any runtime behavior.

One thing in particular to call out: MQ took advantage of JS/PJs flexible prototype-based inheritance model to allow extending the definition of classes like `Controller` over several different files by having each file add more properties to a prototype. This is nice for separating different concerns into different files, but makes it hard for TS to infer correct types.

To deal with this problem, we changed classes that were extended over multiple files to chains of subclasses, each defined in a different file. The final class is then the last subclass in the chain.

This works, but it doesn't seem as elegant as MQ's original version. We may want to consider later refactors to remove some of these classes, perhaps using a "composition over inheritance" approach.


src/shared_types.d.ts
```diff
@@ -0,0 +1,121 @@
+type NodeRef = MQNode | 0;
```
jaltekruse, line 1: I know that this typescript conversion was trying to preserve most of the structure of the original codebase, but handling of null/undefined I think is something I'd like to discuss going forward. I don't know why 0 was chosen as a special value originally instead of just using undefined or null when a NodeRef isn't set.

I've been trying to see what the typescript pattern might exist that would accomplish something like Scala's Option type. It's not a perfect solution to null pointers, but it does give an explicit compile hint that accessors of a property/object should be handling a potential missing value.

https://www.scala-lang.org/api/2.13.6/scala/Option.html

jwmerrill:
I suspect that MQ’s use of 0 in weird places is related to its decision to
define L and R as -1 and 1 so that you can use negation to swap them. There
are places where this is extended to using 0 to represent something like
“undirected” so that negation swaps directions and leaves “undirected”
unchanged.

This doesn’t directly imply that you need to use 0 to represent missing
values in lots of other places, but once you’re doing it somewhere, it
makes it a little less strange to do it in lots of places.

In a future refactor, I’d be interested in getting rid of the negation for
swapping idiom. It’s no longer a shortcut in a typed world because you need
to assert everywhere that the result is a `Direction` and not a generic
number. I’d replace it with an `opp` function that can be typed more
precisely.

Once you do that, 0, 1, and -1 no longer seem so special. Maybe we’d leave
them alone, maybe we’d change them. But I think it would definitely make
sense to remove the use of 0 in places other than directions (and likely
replace it with `undefined`).

src/commands/math.ts
```diff
@@ -42,131 +47,147 @@ var MathElement = P(Node, function(_, super_) {
   };
   // Remove nodes that are more than `cutoff`
   // blocks deep from this node.
-  _.removeNodesDeeperThan = function (cutoff) {
+  removeNodesDeeperThan (cutoff:number) {
     var depth = 0;
-    var queue = [[this, depth]];
-    var current;
+    var queue:[[MQNode, number]] = [[this, depth]];
```
jaltekruse, line 52: I believe this type for the queue should be declared an array of tuples, rather than an tuple of tuples.

Tried learning why this code even works i.e. pushing and shifting from a tuple. Tuples are described in many places as “fixed length ordered collections”. My understanding is that this type definition should be declaring the outer tuple only ever has a single element. Apparently you can also treat a tuple like an array for methods like push/pop/shift and iterating over it, which makes this work but I assume it is discouraged as trying to reference any element past the 0th, with something like queue[1] will fail.

https://dev.to/johan_du_toit/typescript-tuples-are-just-fixed-length-arrays-449n

People have proposed fixing this in the language, but it would be a breaking change. Rather annoyingly people also say all over the place “tuples are just arrays”, which doesn’t make any sense to me. It’s a different type because it existing as a separate type is useful, and it would be nice if it just had consistent behavior.
https://stackoverflow.com/a/64070080/1387961
https://github.com/microsoft/TypeScript/issues/6325

src/commands/math.ts
```diff
@@ -42,131 +47,147 @@ var MathElement = P(Node, function(_, super_) {
...
+    var current:[MQNode, number] | undefined;
 
     // Do a breadth-first search of this node's descendants
     // down to cutoff, removing anything deeper.
-    while (queue.length) {
-      current = queue.shift();
-      current[0].children().each(function (child) {
+    while (current = queue.shift()) {
+      var c = current;
```
jaltekruse, line 58: Its not a big deal wither way, but @mikehaverstock was there a reason I'm missing for why we need this new variable c?

src/commands/math.ts
```diff
@@ -307,19 +331,20 @@ var MathCommand = P(MathElement, function(_, super_) {
...
+      var num1 = parseInt($1, 10);
+      return ' mathquill-block-id=' + blocks[num1].id + ' aria-hidden="true">' + blocks[num1].join('html');
     });
   };
 
   // methods to export a string representation of the math tree
-  _.latex = function() {
-    return this.foldChildren(this.ctrlSeq, function(latex, child) {
+  latex () {
+    return this.foldChildren(this.ctrlSeq || '', function(latex, child) {
```
jaltekruse, line 342: ctrlSeq || ''   appears often enough throughout this changeset that we probably want to consider making ctrlSeq accessed by a function/method that does this coalescing rather than requiring so many references spread around to remember to add this.

src/commands/math/LatexCommandInput.ts
```diff
@@ -2,102 +2,118 @@
...
+    super.createLeftOf(cursor);
 
     if (this._replacedFragment) {
       var el = this.jQ[0];
       this.jQ =
         this._replacedFragment.jQ.addClass('mq-blur').bind(
           'mousedown mousemove', //FIXME: is monkey-patching the mousedown and mousemove handlers the right way to do this?
           function(e) {
-            $(e.target = el).trigger(e);
+            // TODO - overwritting e.target
```
jaltekruse, line 72: Leftover TODO - can we open a ticket to track this?

src/commands/math/advancedSymbols.ts
```diff
@@ -0,0 +1,339 @@
+/************************************
+ * Symbols for Advanced Mathematics
+ ***********************************/
+
+LatexCmds.notin =
+LatexCmds.cong =
+LatexCmds.equiv =
+LatexCmds.oplus =
+LatexCmds.otimes = (latex:string) => new BinaryOperator('\\'+latex+' ', '&'+latex+';');
```
jaltekruse, line 9: I know this isn't particularly specific to this changeset, as the previous code behaved similarly, but the bindBinaryOperator method returns a function that takes no parameters and produces a BinaryOperator. It's a little weird that either of these signatures is fine to put into the LatexCmds map/object.

Also the general pattern of hiding data inside of these "factory methods" caused me some grief trying to write code that applied to all of the operators. In particular this pattern of defining totally different symbols with the same line of code and then relying on later code to pass back in the right latex made determining the information I needed seemingly impossible from where I was working. I ended up making a minor refactor here that I think makes the code clearer, and avoids this overloading of the LatexCmds data structure hold both paremeterless functions and those that require passing in the latex. https://github.com/mathquill/mathquill/pull/960/files#diff-a4c52f3aefd5adb5415b14a7e081bc82f379b5bf64b7fe889794beea20534b83R5

src/commands/math.ts
```diff
@@ -330,100 +355,122 @@ var MathCommand = P(MathElement, function(_, super_) {
...
+  constructor (ctrlSeq?:string, html?:string, text?:string, mathspeak?:string, treatLikeSymbol?:boolean) {
+    if (treatLikeSymbol) {
+      super(ctrlSeq, '<span>'+(html || ctrlSeq)+'</span>', undefined, mathspeak);
+    } else {
+      super(ctrlSeq, '<span class="mq-binary-operator">'+html+'</span>', text, mathspeak);
+    }
   };
-});
+};
+function bindBinaryOperator (ctrlSeq?:string, html?:string, text?:string, mathspeak?:string) {
```
jaltekruse, line 437: I think this function could maybe use a better name now that we are getting rid of the generic unsafe “bind” concept

src/commands/math/basicSymbols.ts
```diff
@@ -53,19 +57,19 @@ var DigitGroupingChar = P(Symbol, function(_, super_) {
     }
 
     // trim the leading spaces
-    while (right !== left && left.ctrlSeq === SPACE) {
+    while (right !== left && left && left.ctrlSeq === SPACE) {
```
jaltekruse, line 60: Minor, could simplify to left?.ctrlSeq

src/commands/math/basicSymbols.ts
```diff
@@ -200,22 +227,22 @@ var Variable = P(Symbol, function(_, super_) {
     } else {
       if (this[L] && !(this[L] instanceof Variable)
           && !(this[L] instanceof BinaryOperator)
-          && this[L].ctrlSeq !== '\\ ')
+          && (this[L] as MQNode).ctrlSeq !== '\\ ')
```
jaltekruse, line 230: Minor note, even if we made the coalescing behavior `.ctrlSeq || ''` happen unconditionally when we access the property as I suggested above, it wouldn't break this usage that doesn't do it (because it obviously doesn't do anything useful here)

src/commands/math/basicSymbols.ts
```diff
@@ -266,33 +303,54 @@ optionProcessors.autoParenthesizedFunctions = function (cmds) {
...
+          l = this;
+          for (i = 1; l && i < str.length; i += 1, l = l[L]);
+
+          new Fragment(l as MQNode, this).remove();
+          cursor[L] = (l as MQNode)[L];
+
+          var cmd = LatexCmds[str];
+          var node;
+          if (isMQNodeClass(cmd)) {
+            node = new (cmd as typeof TempSingleCharNode)(str); // TODO - How do we know that this class expects a single str input?
```
jaltekruse, line 341: @mikehaverstock thoughts on this remaining todo, should we open a ticket to track?

src/commands/math/basicSymbols.ts
```diff
@@ -357,30 +423,50 @@ var Letter = P(Variable, function(_, super_) {
...
+    let autoOpsLength = autoOps._maxLength || 0;
+    
     // check for operator names: at each position from left to right, check
     // substrings from longest to shortest
-    outer: for (var i = 0, first = l[R] || this.parent.ends[L]; i < str.length; i += 1, first = first[R]) {
-      for (var len = min(autoOps._maxLength, str.length - i); len > 0; len -= 1) {
+    outer: for (var i = 0, first = (l as MQNode)[R] || this.parent.ends[L]; first && i < str.length; i += 1, first = (first as MQNode)[R]) {
+      for (var len = min(autoOpsLength, str.length - i); len > 0; len -= 1) {
         var word = str.slice(i, i + len);
+        var last:MQNode = undefined!; // TODO - TS complaining that we use last before assigning to it
```
jaltekruse, line 444: I'm guessing this got resolved? I didn't see any typescript lint errors on the top of this branch.

src/commands/math/basicSymbols.ts
```diff
@@ -487,30 +575,33 @@ optionProcessors.autoOperatorNames = function(cmds) {
   dict._maxLength = maxLength;
   return dict;
 };
-var OperatorName = P(Symbol, function(_, super_) {
-  _.init = function(fn) { this.ctrlSeq = fn; };
-  _.createLeftOf = function(cursor) {
+class OperatorName extends MQSymbol {
+  ctrlSeq:string;
+  constructor (fn?:string) {
+    super(fn || '');
```
jaltekruse, line 581: Is there any need to do this instead of giving a default value in the method signature?

src/commands/math/basicSymbols.ts
```diff
@@ -812,175 +909,189 @@ function isBinaryOperator(node) {
...
+    super('+', '+');
   };
-  _.mathspeak = function() {
+  mathspeak ():string {
     return isBinaryOperator(this) ? 'plus' : 'positive';
   };
-});
+};
 
 //yes, these are different dashes, en-dash, em-dash, unicode minus, actual dash
```
jaltekruse, line 940: This comment could be moved down past the new small class.

src/commands/math/commands.ts
```diff
@@ -285,127 +304,137 @@ var SupSub = P(MathCommand, function(_, super_) {
...
         }
         cmd.supsub = oppositeSupsub;
         delete cmd[supsub];
-        delete cmd[updown+'Into'];
-        cmd[oppositeSupsub][updown+'OutOf'] = insLeftOfMeUnlessAtEnd;
-        delete cmd[oppositeSupsub].deleteOutOf;
+        delete cmd[`${updown}Into`];
+        const cmdOppositeSupsub = cmd[oppositeSupsub]!;
+        cmdOppositeSupsub[`${updown}OutOf`] = insLeftOfMeUnlessAtEnd;
+        delete (cmdOppositeSupsub as any).deleteOutOf; // TODO - refactor so this method can be optional
```
jaltekruse, line 381: Leftover todo, possibly file a follow up ticket

src/commands/text.ts
```diff
@@ -0,0 +1,470 @@
...
+    super.createLeftOf(cursor);
+
+    cursor.insAtRightEnd(textBlock);
+
+    if (textBlock.replacedText)
+      for (var i = 0; i < textBlock.replacedText.length; i += 1)
+        textBlock.write(cursor, textBlock.replacedText.charAt(i));
+
+    const textBlockR = textBlock[R];
+    if (textBlockR) textBlockR.siblingCreated(cursor.options, L);
```
jaltekruse, line 47:
```
 37 ⋮    │-    if (textBlock[R].siblingCreated) textBlock[R].siblingCreated(cursor.options, L);
 38 ⋮    │-    if (textBlock[L].siblingCreated) textBlock[L].siblingCreated(cursor.options, R);
 39 ⋮    │-    textBlock.bubble(function (node) { node.reflow(); });
    ⋮ 46 │+    const textBlockR = textBlock[R];
    ⋮ 47 │+    if (textBlockR) textBlockR.siblingCreated(cursor.options, L);
    ⋮ 48 │+    const textBlockL = textBlock[L];
    ⋮ 49 │+    if (textBlockL) textBlockL.siblingCreated(cursor.options, R);
    ⋮ 50 │+    textBlock.bubble(function (node) { node.reflow(); return undefined});
 40 ⋮ 51 │   };
```
The diff here makes me curious if we are sure that the siblingCreated method will always be defined for this class/node. It seems like if the old code was remotely safe we shouldn't need this check. It would seem odd to make the particular check for a method being present in the old code. Any thoughts @laughinghan ?

src/commands/text.ts
```diff
@@ -0,0 +1,470 @@
...
+class RootMathCommand extends MathCommand {
+  cursor:Cursor;
+  constructor (cursor:Cursor) {
+    super('$');
+    this.cursor = cursor;
+  };
+  htmlTemplate = '<span class="mq-math-mode">&0</span>';
+  createBlocks () {
+    super.createBlocks()
+    const endsL = this.ends[L] as RootMathCommand; // TODO - how do we know this is a RootMathCommand?
```
jaltekruse, line 415: Lingering new todo, open a follow up ticket?

src/commands/text.ts
```diff
@@ -0,0 +1,470 @@
...
+    endsL.cursor = this.cursor;
+    endsL.write = function(cursor:Cursor, ch:string) {
+      if (ch !== '$')
+        MathBlock.prototype.write.call(this, cursor, ch);
+      else if (this.isEmpty()) {
+        cursor.insRightOf(this.parent);
+        this.parent.deleteTowards(undefined!, cursor);
```
jaltekruse, line 422: What does undefined! mean?

src/cursor.ts
```diff
@@ -113,77 +139,87 @@ var Cursor = P(Point, function(_) {
...
+  unwrapGramp () {
     var gramp = this.parent.parent;
     var greatgramp = gramp.parent;
     var rightward = gramp[R];
     var cursor = this;
 
     var leftward = gramp[L];
     gramp.disown().eachChild(function(uncle) {
-      if (uncle.isEmpty()) return;
+      if (uncle.isEmpty()) return true;
```
jaltekruse, line 150: It doesn't look like this return value is used by the one caller I see, should this be declared a void method?

src/tree.ts
```diff
@@ -0,0 +1,560 @@
...
+
+function foldNodes <T>(ends:Ends, fold:T, yield_:(fold:T, el:MQNode) => T):T {
+  var el = ends[L];
+  if (!el) return fold;
+
+  var stop = ends[R];
+  if (!stop) return fold; //shouldn't happen because ends[L] is defined;
+  stop = stop[R];
+
+  // TODO - this cast as MQNode is actually important to keep tests passing. I went to
```
jaltekruse, line 106: New TODO, would be good to open a ticket

src/tree.ts
```diff
@@ -0,0 +1,560 @@
...
+      setTimeout(NodeBase.cleanDictionary);
+    }
+  }
+  static cleanDictionary () {
+    NodeBase.cleaningScheduled = false;
+    NodeBase.TempByIdDict = {};
+  }
+
+
+  // TODO - life would be so much better in typescript of these were undefined instead of
```
jaltekruse, line 172: +1

src/tree.ts
```diff
@@ -0,0 +1,560 @@
...
+    return self;
+  };
+
+  remove () {
+    this.jQ.remove();
+    return this.disown();
+  };
+
+  each (yield_:(el:MQNode) => boolean | undefined) {
+    eachNode(this.ends as Ends, yield_); // TODO - the types of Ends are not compatible
```
jaltekruse, line 539: Lingering TODOs, please file a follow up ticket

jaltekruse (commented):
+1 Thank you @mikehaverstock and @jwmerrill for the great work on this!

src/commands/math/basicSymbols.ts
jaltekruse: Nevermind, I just saw the trailing bang and looked it up, you can ignore my other question about this syntax as well.

https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

src/commands/text.ts
jaltekruse: As noted above, I just learned what this is for in typescript.

https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

src/commands/math/basicSymbols.ts
kschaefe: Telling TypeScript that your `undefined` value is, in fact, defined seems like bad practice.  Why not fix this by changing the type of `last` to `MQNode | undefined`?  If you're putting an `undefined` in there, you better be checking for it.

src/commands/text.ts
kschaefe: As I just mentioned, this seems like bad practice.  If you're going to pass `undefined` into the method, then adjust the parameters for the method to take `undefined` and make sure that method properly checks for it.

src/publicapi.ts
```diff
@@ -67,50 +118,57 @@ function getInterface(v) {
    *   assert(MQ(mathFieldSpan).id === MQ(mathFieldSpan).id);
    *
    */
-  function MQ(el) {
+  var MQ:MQ = function (el:HTMLElement) {
     if (!el || !el.nodeType) return null; // check that `el` is a HTML element, using the
       // same technique as jQuery: https://github.com/jquery/jquery/blob/679536ee4b7a92ae64a5f58d90e9cc38c001e807/src/core/init.js#L92
-    var blockNode = Node.getNodeOfElement($(el).children('.mq-root-block')[0]);
+    var blockNode = NodeBase.getNodeOfElement($(el).children('.mq-root-block')[0]) as MathBlock; // TODO - assumng it's a MathBlock
```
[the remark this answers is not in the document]
jwmerrill: I believe this code (and the comment) are removed in the work to remove jQuery.

src/commands/math/advancedSymbols.ts
jwmerrill: Agree that that's a good refactor

src/services/parser.util.ts
```diff
@@ -0,0 +1,220 @@
...
+  // -*- primitive combinators -*- //
+  or <Q>(alternative: Parser<Q>): Parser<T|Q> {
+    pray('or is passed a parser', alternative instanceof Parser);
+
+    var self = this;
+
+    return new Parser(function(stream, onSuccess, onFailure) {
+      return self._(stream, onSuccess, failure);
+
+      function failure(_newStream: string) {
```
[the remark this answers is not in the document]
jwmerrill: Without the `_`, typescript will complain
```
'newStream' is declared but its value is never read. ts(6133)
```

A leading underscore is a convention that tells TS that a parameter is intentionally unused

src/shared_types.d.ts
mikehaverstock: I think 0 was chosen originally because you can attempt to access properties on a number without throwing an error. For instance: https://github.com/desmosinc/mathquill/blob/9de1943220f8d0dcf401ac27e751e9544b4a0c5c/src/services/keystroke.js#L268

This won't throw an error if `cursor[L] === 0` but would throw an error if `cursor[L] === undefined`.

src/commands/math/basicSymbols.ts
mikehaverstock: I think basically all of the `!` and `as MQNode` casts are because the code was already assuming these casts implicitly. The scope of this typescript conversion was to not change behavior or structure. It was to make any implicit assumptions more explicit. We can now find those assumptions easily by grepping.

We had a second project after this that built on this work. A lot of code was restructured to remove these casts. I imagine a few casts still remain. I don't think it now makes sense to try to do much work on top of this PR. We'd already merged it locally and have heavily modified it.

mikehaverstock: It's also possible that I didn't get the exact typings 100% correct on the first pass. It was hard to tell in many cases. I think it would make sense to revisit the casts that remain on top of the work that removes jquery. A lot more time went into honing the types.

src/commands/math/basicSymbols.ts
mikehaverstock: No. Those sound equivalent.

src/commands/math/basicSymbols.ts
kschaefe: Then, that's not the TODO comment, I'd expect for this.  The following seems more on point:
```
// FIXME: asserting `undefined` exists to maintain legacy JS behavior; should be retype as `MQNode | undefined`
```

src/commands/text.ts
mikehaverstock: I think this is one of those cases where `0` works but `undefined` doesn't. I think this was originally checking both:
1) textBlock[R] is defined
2) that block had the siblingCreated method on it.

If `textBlock[R] === 0` we can still look for the method on that number. But it won't exist. So we early return.

There's a chance we added a base noop `siblingCreated` method to all nodes. Then all we needed to check was that the node existed.

src/cursor.ts
mikehaverstock: Possibly. I think this would be fair game to cleanup assuming things didn't change significantly in the PR that removes jquery.

src/shared_types.d.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/981

src/commands/math.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/982

src/commands/math.ts
jwmerrill: I started writing a ticket for this, but now I'm not 100% convinced that it's actually a good idea. I see this idiom show up 5 times in the codebase in functions that build either a latex or text or mathspeak representation of a node.

We could put this pattern behind an accessor, but it's not totally obvious to me that treating an undefined ctrlSeq identically to an empty string ctrlSeq would be the right thing to do in every case.

src/commands/math/basicSymbols.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/984

src/commands/math/commands.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/985

src/tree.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/981

src/tree.ts
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/986

src/services/parser.util.ts
```diff
@@ -0,0 +1,220 @@
...
+    if (arguments.length < 2) max = min;
+    var self = this;
+
+    return new Parser(function(stream, onSuccess, onFailure) {
+      var xs: T[] = [];
+      var result: boolean = true;
+      var failure;
+
+      for (var i = 0; i < min; i += 1) {
+        // TODO, this may be incorrect for parsers that return boolean
```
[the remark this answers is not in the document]
jwmerrill: Tracked as https://github.com/mathquill/mathquill/issues/987

src/services/mouse.ts
```diff
@@ -99,25 +109,24 @@ Controller.open(function(_) {
       ctrlr.seek($(e.target), e.pageX, e.pageY).cursor.startSelection();
 
       rootjQ.mousemove(mousemove);
-      $(e.target.ownerDocument).mousemove(docmousemove).mouseup(mouseup);
+      const anyTarget = e.target as any;  // TODO - why do we need to cast to any?
```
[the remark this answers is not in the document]
jwmerrill: These casts are gone in the latest `desmosinc/main`.

jwmerrill:
I went through and opened tickets based on a lot of the feedback here.

I ran out of steam before opening a ticket for every last new TODO comment. I personally suspect these are more likely to get resolved as we work on the parts of the codebase where they show up, vs. getting resolved because there is a cleanup ticket that points to them.

jwmerrill:
Closed by https://github.com/mathquill/mathquill/pull/1025

PR #979, opened 2022-04-26
WIP: Update tilde behavior

sclower:
There has long been a discrepancy in Mathquill for how a tilde is handled. When typing from the keyboard, we bound the key to write a \sim instead. However, when setting LaTeX via the MQ API or pasting a tilde, Mathquill did not make the transformation and allowed the character straight through-- which also caused it to display incorrectly.

This PR:

- Sets the tilde symbol to belong to the list of LaTeX commands (rather than character commands) so that there is consistency between typing from a keyboard or setting LaTeX externally.
- Adds a new config option called 'interpretTildeAsSim." When false (default), the tilde appears as a nonbreaking space regardless of entry method. When true, tilde characters are transformed to "\sim".


sclower:
Opened this in the wrong repo. Apologies.

PR #991, opened 2022-09-02
Fix CircleCI

laughinghan:
- upgrade to new 2.0 filepath
- upgrade Docker image to "next-gen"
- add ImageMagick dependency
- fix unit testing with custom unit test webdriver script, since Sauce Labs' unit testing API appears to have broken
  + this actually removes more code than it adds because of all the workarounds needed for Sauce Labs' shitty unit testing API
- fix Many-Worlds commit status link
- use Sauce Connect 4.8.1, which writes JSON to the readyfile `~/sauce_is_ready` instead of leaving it empty, so fix our handling of that


jaltekruse:
+1 LGTM

jaltekruse (approved):

PR #1026, opened 2023-09-27
Debug ci

jwmerrill:


jwmerrill:
I think I understand the root of the problem here.

The output of `make server` shows
```
node script/test_server.js
listening on 0.0.0.0:8000
[2023-09-27T20:05:10.745Z]
make test
make[1]: Entering directory '/home/circleci/*********/*********'
mkdir -p ./build
touch ./build/.exists--used_by_Makefile
rm -rf build/fonts
cp -r src/fonts build/fonts
test -e ./node_modules/.installed--used_by_Makefile || rm -rf ./node_modules/ # robust against previous botched npm install
NODE_ENV=development npm install
make: *** [Makefile:163: server] Terminated


Build was canceled
```

So the server step appears to be deleting the node modules folder and then re-installing it. I think this might be racing some of our other work.

jwmerrill:
Eventual solution to the issue I was debugging here is in https://github.com/mathquill/mathquill/pull/1027

PR #1043, opened 2024-01-23
Expose cursor

nedredmond:
Add a `cursor` function to the API to get the cursor from the private controller. We use it a lot in our functions.

`any` type is not ideal, but that's its type in our codebase currently, and getting a real type is a big investment of effort. While trying to stabilize with the new update, let's keep things simple.


nedredmond:
Wrong repository. My apologies.

PR #1025, opened 2023-09-27
Fix selection perf merge

jwmerrill:
Rolls up changes in
* https://github.com/mathquill/mathquill/pull/967
* https://github.com/mathquill/mathquill/pull/968
* https://github.com/mathquill/mathquill/pull/969
* https://github.com/mathquill/mathquill/pull/970

PR #762, opened 2017-07-11
Add basic matrix support

eoghanmcilwaine:
Includes support for parsing latex environments (used by matrix).

It's not currently possible to add columns/rows via the UI. We plan to
add support for this later.

Supported matrix environments:
- matrix
- pmatrix
- bmatrix
- Bmatrix
- vmatrix
- Vmatrix

Example latex:
\begin{bmatrix}x&y\\1&2\end{bmatrix}

PR #1050, opened 2024-10-03
Re-add basic matrix support

FlamedDogo99:
This is a rewrite of #762, which was made obsolete by #1025.

Includes support for parsing latex environments (used by matrix).

Supported matrix environments:

- matrix
- pmatrix
- bmatrix
- Bmatrix
- vmatrix
- Vmatrix

Example latex:
```
\begin{bmatrix}x&y\1&2\end{bmatrix}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules
 /build
+.DS_Store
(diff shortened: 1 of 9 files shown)


FlamedDogo99:
Some clarification: This builds and works fine, however it makes the linter furious. I am not experienced with type script, so any help would be appreciated.

FlamedDogo99:
Closing this in favor of learning typescript first

Issue #1058, opened 2025-06-19
Unexpected behavior with the typedText method when using Tab, Space and Enter characters ('\t', ' ' and '\n')


FFGFlash:
The following code does not work as expected. 
The [docs](https://docs.mathquill.com/en/latest/Api_Methods/#typedtexttext) describe that at least the space character should terminate commands without inserting an escaped space character.
```ts
typedText('\\sqrt\t')
typedText('\\sqrt ')
typedText('\\sqrt\n')
```

## Expected behavior
The tab, space and newline character should finish the command call and move the cursor to the first child block.

## Actual behavior
The tab character inserts a broken tree. 
The space character inserts an escaped space character in the body of the first child block. 
The newline character does nothing.

## Current work around
Currently the only way to achieve this behavior is to run a keystroke call after the typedText
```ts
typedText('\\sqrt')
keystroke('Tab')
typedText('\\sqrt')
keystroke('Space')
typedText('\\sqrt')
keystroke('Enter')
```

PR #1059, opened 2025-06-19
Fix typedText not terminating a latex command with space and tab characters.

FFGFlash:
I noticed that when using typedText, it didn't perform how the [documentation](https://docs.mathquill.com/en/latest/Api_Methods/#typedtexttext) described, instead of space characters finalizing the command and placing the cursor inside the child node, it instead was finalizing the command and inserting an escaped space character.
```ts
typedText('\\sqrt ') // Fixed
typedText('\\sqrt\t') // Fixed
typedText('\\sqrt\n') // Added conditional check, but due to an earlier if statement this remains unfixed.
```

I also think that a newline character should also terminate the command input, but some earlier code explicitly catches newline characters and calls ``Controller.handle('enter')``.

Related issue: #1058

### DIFF
--- a/src/commands/math/LatexCommandInput.ts
+++ b/src/commands/math/LatexCommandInput.ts
@@ -48,7 +48,7 @@ CharCmds['\\'] = class LatexCommandInput extends MathCommand {
         var cmd = (this.parent as LatexCommandInput).renderCommand(cursor);
         // TODO needs tests
         cursor.controller.aria.queue(cmd.mathspeak({ createdLeftOf: cursor }));
-        if (ch !== '\\' || !this.isEmpty()) cursor.parent.write(cursor, ch);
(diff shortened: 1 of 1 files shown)


drlippman:
I know you wrote this ages ago so this probably won't be useful, but I was playing around with this and it still didn't seem to exactly mimic the behavior of typing.  I was able to get better behavior by modifying the `typedText` function in `textarea.ts` and replacing the `cursor.parent.write(cursor,ch);` line with:
```
if (ch === '\t') {
  cursor.controller.keystroke('Tab');
} else if (ch === ' ') {
  cursor.controller.keystroke('Spacebar');
} else {
  cursor.parent.write(cursor, ch);
}
```
That way `"\t"` and `" "` are handled like the Tab or Spacebar keys were pressed, rather than trying to handle them in `.write()`.

